PR #1, opened 2022-11-08
Add BLOOM image

harubaru:
Adds a Dockerfile that sets up an environment for BLOOM training

Fixes https://github.com/coreweave/infra-pm/issues/332


salanki:
This looks clean, but regretfully I should have probably clarified the requirements a bit since everything outside of  conda pretty much lives outside of the Dockerfile.

Do these changes:
1. For FROM, use our [nccl-tests](https://github.com/coreweave/nccl-tests) image for cuda 11.7, use this artifact: `coreweave/nccl-tests:2022-11-06_19-21-22.11_EST`.
2. Remove row 3 to 70.
3. Then follow the instructions at the bottom off https://github.com/bigscience-workshop/bigscience/tree/master/train/tr11-176B-ml#environment-setup "environment setup", ie:
4. create and activate the conda env (might need a conda init first to get the shell prepared for conda)
5. install transformers
6. install pytorch with `conda install pytorch torchvision torchaudio pytorch-cuda=11.7 -c pytorch -c nvidia`
7. `pip install deepspeed` and follow the rest of the instructions, you will need to find the right deepspeed repo and megatron-deepspeed repo to clone

bloom/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+    rm Miniconda3-latest-Linux-x86_64.sh
+
+ENV PATH=/opt/conda/bin:$PATH
+
+RUN conda create -y -n tr11-176B-ml python=3.8
+
+SHELL ["conda", "run", "-n", "tr11-176B-ml", "/bin/bash", "-c"]
+
+RUN pip install transformers && \
+    pip install --pre torch torchvision torchaudio -f https://download.pytorch.org/whl/test/cu115/torch_test.html -U
```
rtalaricw, line 19: @harubaru You might want to match the CUDA versions (11.7 or 11.6) with Torch CUDA

rtalaricw: `conda install pytorch torchvision torchaudio pytorch-cuda=11.7 -c pytorch -c nvidia`

harubaru: Is there any possibility to use `pip` instead? Using conda to install pytorch breaks the step for installing apex

rtalaricw: Yup that works.

bloom/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN apt-get install -y --no-install-recommends \
+    python3 python3-dev python3-pip git libssl-dev pkg-config
+
+RUN wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh && \
+    bash Miniconda3-latest-Linux-x86_64.sh -b -p /opt/conda && \
+    rm Miniconda3-latest-Linux-x86_64.sh
+
+ENV PATH=/opt/conda/bin:$PATH
+
+# create conda environment from file. this step installs torch.
```
rtalaricw, line 14: I was just curious where torch is getting installed since I do not see it in the environment.yaml file. If you can please explain how it gets installed, it LGTM.

harubaru: Torch and Transformers get installed as a dependency for the packages used in the environment, as installing the packages manually in a separate line creates a ton of package conflict errors.

bloom/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+COPY ./environment.yaml /opt/nccl-tests/environment.yaml
+RUN conda env create -f environment.yaml
+
+SHELL ["conda", "run", "-n", "tr11-176B-ml", "/bin/bash", "-c"]
```
salanki, line 18: Didn't know about this Dockerfile option actually, neat!

bloom/Dockerfile
harubaru: Yep, using their conda resulted in a ton of conflict errors, but it seems like leaving the torch and cudatoolkit dependencies out fixes the issue? Only downside of course is that torch version can't be explicitly specified

harubaru:
Only issue is finding out where the repositories should sit inside of the image which was why I didn't originally include them, but running a git clone inside of the image should suffice, maybe directory structure should be setup in the image as well? Only reason why I didn't include cloning and installing the dependencies in the megatron repo was that the conda env had all of the dependencies for megatron anyway

BigScience is pulled here: https://github.com/jagwar/customAMI-benchmark/blob/bloom-AMI/create-conda-env.sh#L30
Megatron is pulled here also: https://github.com/jagwar/customAMI-benchmark/blob/bloom-AMI/create-conda-env.sh#L24

wbrown:
@salanki is this ready for merge?

salanki:
Almost. I have been using the branch myself. Have some changes I might make so let's leave it open.

wbrown:
@salanki Still pending? Let's not leave this PR open forever. :)

PR #8, opened 2023-01-07
Tensorizer serialization image for Stable Diffusion inference example

harubaru:


wbrown (approved):
Looks better to me. There's going to be a new torch-base shortly.

wbrown (approved):

PR #19, opened 2023-04-13
feat(torch): Add torchaudio, cuDNN support for NCCL builds, and PNG/JPEG support for torchvision

Eta0:
# New Libraries
- This PR adds the `torchaudio` PyTorch extension library to the `torch` container build process.
- The other main PyTorch extension library, `torchvision`, was already included, but its build has been updated as well to include proper PNG/JPEG file support.
- Since the `nccl-tests` base containers now include cuDNN, `torch:nccl` builds now explicitly compile PyTorch and its extensions with cuDNN support enabled.
- A configurable compiler cache was also added in the build stage, to make the builds take slightly less of forever.
  - A *15* GiB cache cuts around an hour off of build times in my testing. PyTorch recommends 25 GiB.
  - It isn't known how well the GitHub Actions runner will handle it (if at all), so it's being added here starting at 1 GiB.
    - This is adjustable with a `docker build` argument, so it can be increased or decreased fairly easily later on.

### DIFF
--- a/.github/workflows/torch-base.yml
+++ b/.github/workflows/torch-base.yml
@@ -16,11 +16,13 @@ jobs:
         include:
           - torch: 2.0.0
             vision: 0.15.1
+            audio: 2.0.1
 
     uses: ./.github/workflows/torch.yml
     with:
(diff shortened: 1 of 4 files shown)


wbrown (approved):
Great job. I liked the DRY'ing up of some of the env vars. :)

PR #20, opened 2023-05-16
feat(cuda-ssh): Add CUDA SSH server container

Eta0:
# CUDA SSH Server
This adds a container for a simple SSH server in a pod corresponding to [this documentation example](https://docs.coreweave.com/coreweave-kubernetes/examples/cuda-ssh).

Used in: [`coreweave/kubernetes-cloud/cuda-ssh`](https://github.com/coreweave/kubernetes-cloud/tree/master/cuda-ssh)
Adapted from: [docker.io/atlanticcrypto/cuda-ssh-server](https://hub.docker.com/r/atlanticcrypto/cuda-ssh-server/tags)

Its base image is `ghcr.io/coreweave/ml-containers/torch`, from this repository.
At the moment it uses the `torch:base` edition because `torch:nccl` is insuitable for SSH sessions, as its necessary environment variables don't propagate into SSH sessions. This may be switched to use `torch:nccl` as a base pending an update upstream in the [`coreweave/nccl-tests`](https://github.com/coreweave/nccl-tests) repo and its images to fix that issue.

### DIFF
--- a/.github/workflows/cuda-ssh.yml
+++ b/.github/workflows/cuda-ssh.yml
@@ -1,0 +1,17 @@
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - "cuda-ssh/**"
+      - ".github/workflows/cuda-ssh.yml"
+      - ".github/workflows/build.yml"
+
+
+jobs:
+  build:
+    uses: ./.github/workflows/build.yml
+    with:
+      image-name: cuda-ssh
+      folder: cuda-ssh
+      build-args: |
+        BASE_IMAGE=ghcr.io/coreweave/ml-containers/torch:bb02bee-base-cuda11.8.0-torch2.0.0-vision0.15.1-audio2.0.1
--- a/.github/workflows/cuda-ssh.yml
+++ b/.github/workflows/cuda-ssh.yml
@@ -9,9 +9,16 @@ on:
 
 jobs:
   build:
+    strategy:
+      matrix:
+        tag:
+          - ceeb8c2-base-cuda11.8.0-torch2.0.1-vision0.15.2-audio2.0.2
+          - ceeb8c2-nccl-cuda11.8.0-nccl2.16.2-1-torch2.0.1-vision0.15.2-audio2.0.2
+
     uses: ./.github/workflows/build.yml
     with:
       image-name: cuda-ssh
       folder: cuda-ssh
+      tag-suffix: torch-${{ matrix.tag }}
       build-args: |
-        BASE_IMAGE=ghcr.io/coreweave/ml-containers/torch:bb02bee-base-cuda11.8.0-torch2.0.0-vision0.15.1-audio2.0.1
+        BASE_IMAGE=ghcr.io/coreweave/ml-containers/torch:${{ matrix.tag }}
--- a/cuda-ssh/Dockerfile
+++ b/cuda-ssh/Dockerfile
@@ -1,0 +1,48 @@
+ARG BASE_IMAGE="ghcr.io/coreweave/ml-containers/torch:bb02bee-base-cuda11.8.0-torch2.0.0-vision0.15.1-audio2.0.1"
+
+FROM ${BASE_IMAGE}
+
+RUN apt-get -qq update && \
+    DEBIAN_FRONTEND=noninteractive \
+      apt-get -qq install --no-install-recommends -y \
+      # Critical packages:
+      ssh ca-certificates tini bash \
+      # Helpful packages:
+      libncurses5 curl wget sudo htop git rsync \
+      tmux unzip nano vim apt-utils iputils-ping && \
+    apt-get clean && \
+    # Wipe the server-side SSH keys on the container image level
+    # to prevent leaking the private host keys, which could
+    # potentially allow impersonation of the SSH server by an attacker.
+    rm /etc/ssh/ssh_host_*
+
+# Since there are no host keys, the SSH server
+# MUST be configured at runtime by running:
+#   dpkg-reconfigure openssh-server
+# (Or by adding custom host key files to /etc/ssh/) before launching it with:
+#   service ssh start
+# Or (blocking):
+#   service ssh start -D
+
+RUN \
+    # Configure the privilege separation directory for sshd
+    # See here for details: https://github.com/openssh/openssh-portable/blob/master/README.privsep
+    install -d --mode=0755 --owner=0 --group=0 /var/run/sshd && \
+    # Configure an empty authorized keys file with correct permissions
+    install -d --mode=0700 --owner=0 --group=0 /root/.ssh && \
+    install --mode=600 --owner=0 --group=0 /dev/null /root/.ssh/authorized_keys && \
+    # Allow only public key authentication
+    install --mode=600 --owner=0 --group=0 /dev/null /etc/ssh/sshd_config.d/10-key-auth.conf && \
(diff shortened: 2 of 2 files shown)


cuda-ssh/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+RUN \
+    # Configure the privilege separation directory for sshd
+    # See here for details: https://github.com/openssh/openssh-portable/blob/master/README.privsep
+    install -d --mode=0755 --owner=0 --group=0 /var/run/sshd && \
+    # Configure an empty authorized keys file with correct permissions
+    install -d --mode=0700 --owner=0 --group=0 /root/.ssh && \
+    install --mode=600 --owner=0 --group=0 /dev/null /root/.ssh/authorized_keys && \
+    # Allow only public key authentication
+    install --mode=600 --owner=0 --group=0 /dev/null /etc/ssh/sshd_config.d/10-key-auth.conf && \
+    echo "PasswordAuthentication no" >> /etc/ssh/sshd_config.d/10-key-auth.conf && \
```
rtalaricw, line 41: Just curious if we make this a requirement to only use SSH Public Key or give them the option to use password?

rtalaricw (approved):
Minor NIT, LGTM

arsenetar:
Is the plan to update the documentation & example files to use this image for the deployment as well as to initialize the image following a similar manner to what had been done previously?  Using key based authentication seems like a safer default, although documentation for the example does follow password auth.

Eta0:
> Is the plan to update the documentation & example files to use this image for the deployment as well as to initialize the image following a similar manner to what had been done previously? Using key based authentication seems like a safer default, although documentation for the example does follow password auth.

@arsenetar Yes, I was going to suggest a new series of steps for the docs in the corresponding PR for `kubernetes-cloud`, but I couldn't make that PR earlier because GitHub went down. The new steps would be:

```bash
# Create persistent storage volumes
kubectl apply -f sshd-root-pvc.yaml -f sshd-data-pvc.yaml

# Create the SSH service and public IP
kubectl apply -f sshd-service.yaml

# Launch the SSH server
kubectl apply -f sshd-deployment.yaml

# Wait for the SSH server to come up
kubectl rollout status deployment/sshd

# Add your SSH public key
kubectl exec -i deployment/sshd -c sshd -- /bin/tee --append /root/.ssh/authorized_keys < ~/.ssh/id_rsa.pub

# Print the public IP address
kubectl get service/sshd -o jsonpath='service/sshd public IP: {.status.loadBalancer.ingress[0].ip}{"\n"}'

# Print the host key fingerprints
kubectl exec deployment/sshd -c sshd -- find /etc/ssh -type f -name 'ssh_host_*_key.pub' -exec ssh-keygen -lf {} ';'

# Connect
ssh root@<public IP>
```

(No preference on keeping the old `kubectl get service` versus the specialized one to print just the IP)
I think the `kubectl exec` line to add a public key ought to be pretty easy to follow.

cuda-ssh/Dockerfile
Eta0: I was thinking we would make public key authentication the only option. If we want to also allow password authentication we could add an alternative configuration file called `11-password-auth.conf` and allow deleting `10-key-auth.conf` at the time of deployment, but the public key authentication setup is pretty straightforward, so I'd lean towards just that.

wbrown: I advocate no passwords, and nor should we encourage our customers to use passwords.

arsenetar:
@wbrown Yeah, just wanted to make sure we were going to update the docs to work with this.

rtalaricw:
@Eta0 We can merge this.

Eta0:
> @Eta0 We can merge this.

@rtalaricw

Nice — I did also want to raise another question before merging. @salanki had suggested using `nccl-tests` as a base for this container, and now that [`nccl-tests`'s PR #16](https://github.com/coreweave/nccl-tests/pull/16) has been merged in, we could actually use `torch:nccl` instead of `torch:base` for these.

Pros of that would be:
- `torch:nccl` is based on the `nvidia/cuda` `devel` containers featuring cuDNN, and contain a lot of CUDA development tools by default, like `nvcc`.
  - If these SSH containers are intended mainly for development, that could make them more usable.

Cons would be:
- `torch:nccl` includes significantly different PyTorch configuration options and supplemental libraries (e.g. HPC-X & newer `nccl` versions), so what works or doesn't work on that type of container may not correspond 1:1 to how a finished aop would work running off of `torch:base` or bare `nvidia/cuda` images.
- `torch:nccl` images are much larger.

What would y'all think fits better between those two options? We could also add in a matrix build for this with both editions.

rtalaricw:
> > @Eta0 We can merge this.
> 
> @rtalaricw
> 
> Nice — I did also want to raise another question before merging. @salanki had suggested using `nccl-tests` as a base for this container, and now that [`nccl-tests`'s PR #16](https://github.com/coreweave/nccl-tests/pull/16) has been merged in, we could actually use `torch:nccl` instead of `torch:base` for these.
> 
> Pros of that would be:
> 
> * `torch:nccl` is based on the `nvidia/cuda` `devel` containers featuring cuDNN, and contain a lot of CUDA development tools by default, like `nvcc`.
>   
>   * If these SSH containers are intended mainly for development, that could make them more usable.
> 
> Cons would be:
> 
> * `torch:nccl` includes significantly different PyTorch configuration options and supplemental libraries (e.g. HPC-X & newer `nccl` versions), so what works or doesn't work on that type of container may not correspond 1:1 to how a finished aop would work running off of `torch:base` or bare `nvidia/cuda` images.
> * `torch:nccl` images are much larger.
> 
> What would y'all think fits better between those two options? We could also add in a matrix build for this with both editions (but it would be sort of complicated to fit a choice between those into the docs example).

Lets do a matrix build! I think keeping it minimal is ideal but providing options is better.

Eta0:
Oops, closed this by mistake. Working on the matrix build now.

Eta0:
Matrix build added, and the [corresponding coreweave/kubernetes-cloud PR](https://github.com/coreweave/kubernetes-cloud/pull/222) is now up as well. It currently uses the version of this container built from `torch:nccl` as its base.

---

> There are reasons to stick with our torch build, including architecture impact and cuBLAS bugs. Image size really does impact things, and we should not underestimate this.

@wbrown either option would use our torch builds; the choice would be between `torch:base` and `torch:nccl`. Since `torch:nccl` is based off of the development version of `nvidia/cuda`, it would probably make for a better development container, but it has the side effects of everything else that comes with `torch:nccl` instead of `torch:base`. Which one do you think works better, in this case?

.github/workflows/cuda-ssh.yml
```diff
@@ -0,0 +1,24 @@
...
+      - ".github/workflows/cuda-ssh.yml"
+      - ".github/workflows/build.yml"
+
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        tag:
+          - ceeb8c2-base-cuda11.8.0-torch2.0.1-vision0.15.2-audio2.0.2
```
wbrown, line 15: This almost belongs in the ml containers repo, as this would let us update to CI. Hm. Not a blocker, though.

wbrown (approved):

wbrown:
@Eta0 `torch-nccl`, quite reluctantly despite the increase in size. This would let the image work on more machines and architectures.

PR #26, opened 2023-06-29
feat: PyTorch Extras Container

Eta0:
# `torch-extras` Container

This PR adds a new container named `ml-containers/torch-extras`, which is `ml-containers/torch` with supplementary libraries [DeepSpeed](https://github.com/microsoft/DeepSpeed) and [flash-attention](https://github.com/HazyResearch/flash-attention).
The code is originally based off of #21, but significantly more generalized and with the finetuner application-specific parts removed.

## Rationale

DeepSpeed and flash-attention both require CUDA development tools to install properly. This complicates using them with anything but an `nvidia/cuda:...-devel` based image. Optionally including them with our `ml-containers/torch` containers allows for still-lightweight images that can use those powerful libraries without the full CUDA development toolkit. It also reduces compile time for downstream Dockerfiles, since flash-attention takes a long time to compile at whatever step it is included.

## Structure

`ml-containers/torch-extra` is separated out as a separate container, unlike the tag-differentiated `torch:base` and `torch:nccl` flavours of the baseline torch image. These are simply layers on top of the `torch:base` and `torch:nccl` images, and are built as a second CI step immediately after either of those two are built.
Since compatibility of DeepSpeed and flash-attention may lag behind PyTorch releases themselves, the secondary step to build these images can be temporarily disabled via flags in `torch-base.yml` and `torch-nccl.yml` until they become compatible.

I welcome comments and suggestions on this build process and structure, because it requires tradeoffs. It guarantees that the `torch-extras` containers are always built, whenever possible, on new `torch` image updates, but it makes it more difficult to build the `torch-extras` containers standalone, if desired.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,11 +15,25 @@ on:
       tag-suffix:
         required: false
         type: string
+    outputs:
+      outcome:
+        description: "The outcome of the build"
+        value: ${{ jobs.build.outputs.outcome }}
+      tags:
+        description: "The resulting image tags"
+        value: ${{ jobs.build.outputs.tags }}
+      version:
+        description: "The resulting image version"
+        value: ${{ jobs.build.outputs.tags }}
 
 jobs:
   build:
     name: Build Images
     runs-on: [self-hosted, Linux]
+    outputs:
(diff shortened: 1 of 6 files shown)


wbrown (approved):
:+1:

PR #41, opened 2023-09-24
ci(torch-nightly): Fix DeepSpeed compilation

Eta0:
# Fix DeepSpeed Compilation

This change disables the AIO extension in DeepSpeed from being compiled on torch v2.2 and up, since it is not currently compatible with extension compilation requirements added in torch v2.1.
The AIO extension had previously been disabled only for torch v2.1 in `torch-extras/Dockerfile`, but the latest nightly builds of torch are now onto v2.2, and the issue is still present as of the latest DeepSpeed release.

This also bumps the version of DeepSpeed from v0.10.1 to v0.10.3.


wbrown (approved):
Tested the image, works great.

PR #51, opened 2023-12-20
feat(hf-llm-inference): Add initial HF LLM Inference image

harubaru:
This PR adds a workflow to build the HF LLM Inference image from this PR: https://github.com/coreweave/kubernetes-cloud/pull/303

### DIFF
--- a/.github/workflows/hf-llm-inference.yml
+++ b/.github/workflows/hf-llm-inference.yml
@@ -1,0 +1,23 @@
+on:
+  workflow_dispatch:
+    inputs:
+      commit:
+        description: 'Commit to build'
+        required: true
+        default: 'master'
+  push:
+    paths:
+      - "hf-llm-inference/**"
+      - ".github/workflows/hf-llm-inference.yml"
+      - ".github/workflows/build.yml"
+
+
+jobs:
+  build:
+    uses: ./.github/workflows/build.yml
+    secrets: inherit
+    with:
+      image-name: hf-llm-inference
+      folder: hf-llm-inference
+      build-args: |
+        COMMIT=${{ github.event.inputs.commit }}
--- a/.github/workflows/hf-llm-inference.yml
+++ b/.github/workflows/hf-llm-inference.yml
@@ -4,7 +4,6 @@ on:
       commit:
         description: 'Commit to build'
         required: true
-        default: 'master'
   push:
     paths:
       - "hf-llm-inference/**"
@@ -20,4 +19,4 @@ jobs:
       image-name: hf-llm-inference
       folder: hf-llm-inference
       build-args: |
-        COMMIT=${{ github.event.inputs.commit }}
+        ${{ inputs.commit && 'COMMIT=' }}${{ inputs.commit }}
--- a/hf-llm-inference/Dockerfile
+++ b/hf-llm-inference/Dockerfile
@@ -1,0 +1,21 @@
+FROM ghcr.io/coreweave/ml-containers/torch:afecfe9-base-cuda12.0.1-torch2.0.0-vision0.15.1
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt update && apt upgrade -y && \
+    apt update && apt install -y python3 python3-pip git curl && \
+    apt clean
+
+RUN mkdir /app
+WORKDIR /app
+
+ARG COMMIT=master
+RUN git clone https://github.com/coreweave/kubernetes-cloud && \
+    cd kubernetes-cloud && \
+    git checkout ${COMMIT} && \
+    cd .. && \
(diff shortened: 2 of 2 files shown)


hf-llm-inference/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt update && apt upgrade -y && \
+    apt update && apt install -y python3 python3-pip git curl && \
+    apt clean
+
+RUN mkdir /app
+WORKDIR /app
+
+ARG COMMIT=master
```
Eta0, line 11: Try defaulting to [the commit `4ff78daa65b6a4a944c5f14f102e5f28770bcfcc` from your PR](https://github.com/coreweave/kubernetes-cloud/pull/303/commits/4ff78daa65b6a4a944c5f14f102e5f28770bcfcc) so it can build. Using a branch name as the default leads to weird things because of Dockerfile build layer caching; it will repeatedly build from the same commit as whatever was the `HEAD` on its first run (unless the cache is invalidated or cleared).

hf-llm-inference/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN git clone https://github.com/coreweave/kubernetes-cloud && \
```
Eta0, line 12: Cloning the full git history of `kubernetes-cloud` is excessive because of the weird situation with all the GitBook docs assets being thrown in there until recently; you can make it quicker (hundreds fewer megabytes) by using `--filter=blob:none`. It can be even more selective if you copy [the way I wrote the clone in the `llm-inference` branch](https://github.com/coreweave/ml-containers/blob/d1e0d72c435c20b4554175623a78a68980a73284/llm-finetuner/Dockerfile#L11-L22), but `--filter=blob:none` gets most of it.

```suggestion
RUN git clone --filter=blob:none https://github.com/coreweave/kubernetes-cloud && \
```

hf-llm-inference/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ghcr.io/coreweave/ml-containers/torch:afecfe9-base-cuda12.0.1-torch2.0.0-vision0.15.1
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt update && apt upgrade -y && \
+    apt update && apt install -y python3 python3-pip git curl && \
```
Eta0, line 5: The `torch` base images include Python, and have PyTorch compiled against that particular one. Reinstalling it here may lead to binary compatibility issues.
```suggestion
    apt update && apt install -y git curl && \
```

Eta0 (changes requested):

.github/workflows/hf-llm-inference.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  build:
+    uses: ./.github/workflows/build.yml
+    secrets: inherit
+    with:
+      image-name: hf-llm-inference
+      folder: hf-llm-inference
+      build-args: |
+        COMMIT=${{ github.event.inputs.commit }}
```
Eta0, line 23: This should prevent it from passing `COMMIT=""` on push events
```suggestion
        ${{ inputs.commit && 'COMMIT=' }}${{ inputs.commit }}
```

.github/workflows/hf-llm-inference.yml
```diff
@@ -0,0 +1,23 @@
+on:
+  workflow_dispatch:
+    inputs:
+      commit:
+        description: 'Commit to build'
+        required: true
+        default: 'master'
```
Eta0, line 7: Branch names don't work quite right in Dockerfiles, so better to remove that default
```suggestion
```

wbrown (approved):

hf-llm-inference/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ghcr.io/coreweave/ml-containers/torch:afecfe9-base-cuda12.0.1-torch2.0.0-vision0.15.1
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt update && apt upgrade -y && \
+    apt update && apt install -y git curl && \
+    apt clean
```
Eta0, line 6: Didn't notice the `apt upgrade`, definitely don't want that. That will mess with all sorts of packages.
```suggestion
RUN apt-get -qq update && \
    apt-get -qq install --no-install-recommends -y git curl && \
    apt-get clean
```

Eta0 (changes requested):

Eta0 (approved):

harubaru:
Tested the image with the inference service in the mentioned PR and it works perfectly

PR #57, opened 2024-02-13
fix(vllm-tensorizer): Correct workflow yaml for manual workflows and updates to `vllm-tensorizer` container

sangstar:
This updates the `vllm-tensorizer` Dockerfile with recent fixes and fixes the workflow yaml to track the correct development branch for `vllm-tensorizer`, `ss/vllm-tensorizer`.


vllm-tensorizer/Dockerfile
```diff
@@ -42,8 +49,16 @@ RUN git clone --filter=blob:none --depth 1 --no-single-branch --no-checkout \
 
 FROM builder-base as vllm-builder
 WORKDIR /workspace
-COPY --from=vllm-downloader /git/vllm /workspace
-RUN python3 -m pip wheel -w /wheels -v --no-cache-dir --no-build-isolation --no-deps ./
+ENV LD_LIBRARY_PATH /usr/local/cuda/lib64:${LD_LIBRARY_PATH}
+RUN if [ -d "/usr/local/cuda" ]; then echo "Directory /usr/local/cuda exists."; else exit 1; fi
```
Eta0, line 53: 1. `/usr/local/cuda/lib64` is added to `ld`'s search path by `/etc/ld.so.conf.d/000_cuda.conf` when you run `ldconfig`, so this does not need to be prepended to `LD_LIBRARY_PATH`
2. This check isn't needed (it will almost certainly pass anyway) as long as the following build succeeds
```suggestion
```

vllm-tensorizer/Dockerfile
```diff
@@ -53,12 +68,20 @@ WORKDIR /workspace
...
+      -c /tmp/constraints.txt
 
-# Install the vllm wheel from the vllm-builder layer
 RUN --mount=type=bind,from=vllm-builder,source=/wheels,target=/tmp/wheels \
-    python3 -m pip install --no-cache-dir /tmp/wheels/*.whl
+    python3 -m pip install --no-cache-dir /tmp/wheels/*.whl -c /tmp/constraints.txt && \
+    rm /tmp/constraints.txt
+
+# `dive` shows that the `.cache` folder still exists in the final image
+RUN rm -rf /root/.cache/pip
```
Eta0, line 83: Running `rm -rf` here will not decrease the final image size because container images are copy-on-write (files are only marked as removed, but still take up space). It most likely exists because another `pip` step is missing `--no-cache-dir`, and needs to be fixed there.

```suggestion
```

vllm-tensorizer/Dockerfile
```diff
@@ -53,12 +68,20 @@ WORKDIR /workspace
 
 RUN apt-get -qq update && apt-get install -y --no-install-recommends curl && apt-get clean
 
-RUN pip3 install "transformers == 4.36.0" "fschat[model_worker] == 0.2.30" "triton == 2.1.0"
+RUN --mount=type=bind,from=freezer,target=/tmp/frozen \
+    /tmp/frozen/freeze.sh torch torchaudio torchvision xformers > /tmp/constraints.txt
+
+RUN python3 -m pip install \
+      "fschat[model_worker] == 0.2.30" "triton == 2.1.0" \
+      -c /tmp/constraints.txt
```
Eta0, line 76: This one is missing `--no-cache-dir`
```suggestion
RUN python3 -m pip install --no-cache-dir \
      "fschat[model_worker] == 0.2.30" "triton == 2.1.0" \
      -c /tmp/constraints.txt
```

vllm-tensorizer/Dockerfile
```diff
@@ -53,12 +68,20 @@ WORKDIR /workspace
...
+    python3 -m pip install --no-cache-dir /tmp/wheels/*.whl -c /tmp/constraints.txt && \
+    rm /tmp/constraints.txt
+
+# `dive` shows that the `.cache` folder still exists in the final image
+RUN rm -rf /root/.cache/pip
 
 EXPOSE 8080
 
-ENTRYPOINT ["python3", "-m", "vllm.entrypoints.openai.api_server"]
+#ENTRYPOINT ["python3", "-m", "vllm.entrypoints.openai.api_server"]
```
Eta0, line 87: Double check whether this should be removed or left commented out like this :eyes:

vllm-tensorizer/Dockerfile
```diff
@@ -24,11 +29,13 @@ RUN export \
...
 
-RUN apt update && \
-    apt install -y python3-pip python3-packaging \
-      git ninja-build && \
-    pip3 install -U --no-cache-dir pip packaging setuptools wheel
+RUN ldconfig
 
+RUN apt-get -qq update && \
+    apt-get -qq install -y --no-install-recommends \
+      python3-pip python3-packaging git ninja-build && \
```
Eta0, line 36: The `apt` package `python3-packaging` is equivalent to `pip3 install -U packaging` below

```suggestion
      python3-pip git ninja-build && \
```

Eta0 (changes requested):
Mostly good, can use some cleanup

vllm-tensorizer/Dockerfile
sangstar: No particularly good reason -- sort of a dev comment

Eta0 (approved):

PR #68, opened 2024-05-29
feat(torch): Build with AOCL-BLAS and AOCL-LAPACK, and HPC-X v2.19

Eta0:
# BLAS & LAPACK

This change adds [AOCL-BLAS & AOCL-LAPACK](https://www.amd.com/en/developer/aocl/dense.html) as lightweight AMD-optimized alternatives to MKL for CPU linear algebra. Previously, no BLAS or LAPACK implementations were included.

This additionally updates the `nccl-tests` base images for `torch:nccl` to use HPC-X v2.19, as added in coreweave/nccl-tests#38.

### DIFF
--- a/.github/configurations/torch-nccl.yml
+++ b/.github/configurations/torch-nccl.yml
@@ -3,28 +3,28 @@ image:
   - cuda: 12.2.2
     os: ubuntu22.04
     nccl: 2.19.3-1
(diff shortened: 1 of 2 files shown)


wbrown (approved):
Lordy. :+1:

PR #76, opened 2024-08-21
(feat) Add `TransformerEngine` build.

wbrown:
Adds `TransformerEngine` to `torch-extras`.


wbrown:
@Eta0 nightly seems broken but shouldn’t block this.

PR #79, opened 2024-10-07
feat(megatron): add initial megatron-lm image

harubaru:


.github/workflows/megatron.yml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  build:
+    uses: ./.github/workflows/build.yml
+    secrets: inherit
+    with:
+      image-name: megatron
+      folder: megatron
+      build-args: |
+        BASE_IMAGE=${{ inputs.base-image || 'ghcr.io/coreweave/ml-containers/torch-extras:bfe03aa-nccl-cuda12.4.1-ubuntu22.04-nccl2.21.5-1-torch2.4.0-vision0.19.0-audio2.4.0'}}
```
wbrown, line 22: Do we want to parameterize this?

megatron/requirements.txt
```diff
@@ -0,0 +1,4 @@
+pyyaml==6.0.2
+regex==2024.9.11
+transformers==4.45.2
+triton==3.0.0
```
wbrown, line 4: If this doesn't include megatron, maybe call it `transformers-triton`?

.github/workflows/megatron.yml
harubaru: It seems to be impossible to do this according to https://github.com/orgs/community/discussions/26322

wbrown (approved):

PR #85, opened 2024-10-22
feat(torch): Update `torch` libraries to v2.5.0, bundle `triton`, patch TransformerEngine

Eta0:
# PyTorch v2.5.0, `triton`, & Patched TransformerEngine v1.11

This change includes version updates, some patches, and bundles a source build of `triton` in the non-nightly `ml-containers/torch` images (as they were previously only bundled in the nightly ones).

## PyTorch

The following PyTorch components have been updated:
- `torch` [v2.4.1](https://github.com/pytorch/pytorch/releases/v2.4.1) → [v2.5.0](https://github.com/pytorch/pytorch/releases/v2.5.0)
- `torchvision` [v0.19.1](https://github.com/pytorch/vision/releases/v0.19.1) → [v0.20.0](https://github.com/pytorch/vision/releases/v0.20.0)
- `torchaudio` [v2.4.1](https://github.com/pytorch/audio/releases/v2.4.1) → [v2.5.0](https://github.com/pytorch/audio/releases/v2.5.0)

In addition, the patch to fix `torchaudio` compilation on CUDA 12.5+ (from #82) is now obsolete in the nightly builds of `torchaudio` (so, for some version after v2.5.0), as the contents of the patch were included as a secondary change in pytorch/audio#3843, so this adds a check to the build process to only apply that patch in versions where it is necessary.

## `triton`

`triton` is normally listed as a dependency for the x86_64 Linux releases of PyTorch on PyPI, but it is not automatically registered as a dependency in our images given the way that we build PyTorch from source. Because it would normally be seen as a required dependency, and because each PyTorch release expects a specific version of `triton` to be paired with it (when using features that require `triton`), this PR adds a source build of `triton` as a bundled part of all `ml-containers/torch` images.

The version of `triton` to build is pulled with the same method that `torch-nightly` images use: a commit hash is read from the [`.ci/docker/ci_commit_pins/triton.txt`](https://github.com/pytorch/pytorch/blob/v2.5.0/.ci/docker/ci_commit_pins/triton.txt) file in the cloned `pytorch/pytorch` repository, and that exact git commit of `triton` is used for a build. PyPI releases of `triton` are not used, even though they are generally compatible with stable PyTorch versions built from source, like ours, as we already had the code to build `triton` from a specific commit, and this could potentially be more flexible when building e.g. PyTorch release candidates.

The version of `triton` to be built can still be overridden by setting the `BUILD_TRITON_VERSION` build argument explicitly, and can be turned off completely by setting the new separate `BUILD_TRITON` build argument to any value other than `1`.

## TransformerEngine Updates

This change additionally updates TransformerEngine by a few commits, as it was previously using a commit from right before the v1.11 release. Now that v1.11 is out, it builds from that tag, and additionally includes a patch for a bug in the v1.11 release (NVIDIA/TransformerEngine#1213) that was fixed on the v1.11 branch (and in future versions) but not the v1.11 git tag as part of NVIDIA/TransformerEngine#1222.


wbrown (approved):
:+1:

PR #87, opened 2025-02-14
feat!: Update to new runners, update PyTorch, add ARM builds, add Blackwell support, add `sglang` image

Eta0:
# Many updates (CI, building for more architectures, PyTorch libraries, etc.)

This PR is to merge an overly-long-running branch back to `main` comprising most updates to PyTorch and the CI system from the last few months. Some highlights include:
- Switched to use newer GHA CI runners
- Added support for multi-arch amd64+arm64 image builds (now the default)
  - The `ml-containers/torch` and `ml-containers/torch-extras` build processes were adapted to support arm64, to match
- Dropped Ubuntu 20.04 builds
  - Python 3.8, included in Ubuntu 20.04, is now EOL and unsupported by PyTorch
- Added CUDA 12.8 builds for PyTorch
- Added PyTorch builds with compute capabilities 9.0a, 10.0, and 10.0a enabled (on applicable CUDA versions)
- Added PyTorch builds with the CXX11 ABI disabled, in addition to the ones with it enabled (previous default)
  - This change will likely be reverted later, as PyTorch is planning to migrate to enabling the CXX11 ABI for their own builds, which will mean that keeping it enabled will lead to the best compatibility
- Updated PyTorch (and friends) to v2.6.0
- Updated many `ml-containers/torch` and `ml-containers/torch-extras` bundled libraries
- Added Flash Attention v3 to the `ml-containers/torch` image
- Added an `sglang` image

There were also various bugfixes, improvements, and customizability enhancements added along the way noted in commit messages.


.github/workflows/torch.yml
```diff
@@ -99,7 +97,8 @@ jobs:
         BUILD_TORCH_VERSION=${{ inputs.torch-version }}
         BUILD_TORCH_VISION_VERSION=${{ inputs.torchvision-version }}
         BUILD_TORCH_AUDIO_VERSION=${{ inputs.torchaudio-version }}
-        ${{ inputs.cuda-arch-support && format('BUILD_TORCH_CUDA_ARCH_LIST={0}', inputs.cuda-arch-support) || '' }}
+        BUILD_TORCH_CUDA_ARCH_LIST=7.0 7.5 8.0 8.6 8.9 9.0+PTX
```
wbrown, line 100: Was this intentional to move this out of being parameterized?

wbrown (commented):
Heckin' big. One question, otherwsie looks great to me.

.github/workflows/torch.yml
Eta0: Yes, because we hit the limit of 10 inputs per reusable workflow. In retrospect I think it would be better to pass more of these as a block, like `inputs.additional-build-args`, when they don't need special processing (like the ones that need to be separated out to be included in the image tag).

This one should also have its default moved to the Dockerfile, so that they don't get out of sync.

Eta0: Reworked in 68fbfd11691200e55d920709b9a4802b639b0c29.

wbrown (approved):

PR #88, opened 2025-03-06
feat(s3-perf-test): Create s3-perf-test image

ADustyOldMuffin:
This will run a perf test on a given S3 bucket reading a set amount of objects in 16KiB blocks to simulate a real world workload on an object store.


s3-perf-test/main.go
```diff
@@ -0,0 +1,258 @@
...
+	"log/slog"
+	"math/rand"
+	"os"
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/hashicorp/go-cleanhttp"
+	"github.com/jedib0t/go-pretty/v6/table"
+	"github.com/minio/minio-go/v7"
```
codyohl, line 16: curious: why the minio client?

s3-perf-test/main.go
```diff
@@ -0,0 +1,258 @@
...
+		ObjectSizes: sizes,
+		S3Client:    s3Client,
+	}
+
+	start := time.Now().UTC()
+
+	for i := range threads {
+		wg.Add(1)
+		go func() {
+			result := runTest(ctx, params, i)
```
codyohl, line 186: something that's nice is the ability for all the tests on different nodes to coordinate enough to gang-schedule themselves at approximately the exact same time in the cluster (esp. for various types of workloads)

s3-perf-test/main.go
```diff
@@ -0,0 +1,258 @@
...
+
+			resp, err := params.S3Client.GetObject(testCtx, params.Bucket, obj, reqOpts)
+
+			if err != nil {
+				ll.Error("failed to fetch range", "error", err)
+
+				continue
+			}
+
+			amount, err := io.Copy(io.Discard, resp)
```
codyohl, line 289: if developing write workloads as well, I've found pooling/buffer reuse helpful

s3-perf-test/main.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	transport := cleanhttp.DefaultPooledTransport()
+
+	s3Client, err := minio.New(endpoint, &minio.Options{
+		Creds:        credentials.NewStaticV4(accessKeyID, secretKey, ""),
+		Secure:       ssl,
+		Transport:    transport,
```
codyohl, line 135: curious, why this?

codyohl (commented):
very nice!

s3-perf-test/main.go
ADustyOldMuffin: Far more convenient essentially! No code for endpoint resolvers, pointers, etc. Per is more or less the same since they both use the underlying net/http.

s3-perf-test/main.go
ADustyOldMuffin: Ah sort of like a sync barrier almost?

s3-perf-test/main.go
ADustyOldMuffin: cleanhttp just sets sane defaults on the normal golang net/http. I then took some known good defaults for specifically load testing and then overwrite the defaults that the cleanhttp doesn't set/sets to something I don't want. Essentially its just a quicker short hand, but realistically not needed.

s3-perf-test/main.go
codyohl: > Ah sort of like a sync barrier almost?

yes, precisely that (we use NFS for coordination, since it needs to be low-dep / cluster agnostic ideally)

s3-perf-test/main.go
codyohl: right, makes sense!

s3-perf-test/main.go
codyohl: makes sense, thx!

wbrown:
@ADustyOldMuffin Are we done with this? Merge?

PR #93, opened 2025-05-13
feat(torch): Update to PyTorch 2.7 & CUDA 12.8.1

Eta0:
# PyTorch, CUDA, and NCCL updates

This change updates the following components:
- PyTorch 2.6.0 → 2.7.0
- CUDA 12.8.0 → 12.8.1
- NCCL 2.25.1-1 → 2.26.5-1
Along with a few other miscellaneous component updates.

This also fixes a few build issues (e.g. installing missing runtime libs, pinning the version of CMake used) and improves robustness (e.g. using multiple methods to ensure that `sm_100` is always a targeted arch for builds).


.github/configurations/torch-base.yml
```diff
@@ -1,7 +1,7 @@
-cuda: [ 12.8.0, 12.6.3, 12.4.1 ]
+cuda: [ 12.8.1, 12.6.3, 12.4.1 ]
 os: [ ubuntu22.04 ]
-abi: [ 1, 0 ]
+abi: [ 1 ]
```
sangstar, line 3: Is there any need for us to document us officially no longer supporting `abi0` or is this just implied by updating to PyTorch 2.7?

torch-extras/Dockerfile
```diff
@@ -68,7 +68,10 @@ RUN apt-get -qq update && apt-get -qq install -y \
     { wget -qO - https://apt.kitware.com/keys/kitware-archive-latest.asc \
     | gpg --dearmor -o /etc/apt/trusted.gpg.d/kitware.gpg; } && \
     apt-add-repository "deb https://apt.kitware.com/ubuntu/ $(lsb_release -cs) main" && \
-    apt-get -qq update && apt-get -qq install -y cmake && apt-get clean
+    apt-get -qq update && \
+    apt-get -qq install -y 'cmake=3.31.6-*' 'cmake-data=3.31.6-*' && \
+    apt-get clean && \
+    python3 -m pip install --no-cache-dir 'cmake==3.31.6'
```
sangstar, line 74: Why do we both `pip install` and `apt-get install` `cmake==3.31.6`?

torch/Dockerfile
```diff
@@ -659,6 +662,7 @@ RUN export \
       cuda-nvrtc-${CUDA_PACKAGE_VERSION} \
       libcusparse-${CUDA_PACKAGE_VERSION} \
       libcusolver-${CUDA_PACKAGE_VERSION} \
+      libcufile-${CUDA_PACKAGE_VERSION} \
```
sangstar, line 665: We're downloading another CUDA library? Is this specifically for the updated versions here?

sangstar (approved):
LGTM; couple questions.

torch-extras/Dockerfile
Eta0: No, CMake is not included like this in the built image, only the builder so it wouldn't make sense for it to have an exported version number. It is installed via `pip` as well as `apt-get` because some Python build pipelines try to install it via `pip` anyway without checking if it's already installed at the system level, so pre-installing it with `pip` at the version we want means that those won't break by defaulting to installing the latest version of CMake, because it will find that it's already installed with `pip` and carry on instead.

torch/Dockerfile
Eta0: Yes. `libcufile` is required as a runtime library for the updated software we are using here. The `torch:nccl` images already had this library installed, so this only affects the `torch:base` images.

PR #94, opened 2025-05-16
ci(torch): Build against CUDA 12.9.0, update NCCL base images

Eta0:
# CUDA 12.9.0 & updated NCCL base images

This change adds CUDA 12.9.0 as a build target and updates the `nccl-tests` base images for the `torch:nccl` builds to newer versions that support CUDA 12.9.0 and additionally bundle HPC-X 2.23, up from 2.21.1.


bradbeam (approved):

torch/Dockerfile
```diff
@@ -10,15 +10,15 @@ ARG BUILD_FLASH_ATTN_VERSION="2.7.4.post1"
...
 ARG BUILD_TRANSFORMERENGINE_CUDA_ARCH_LIST="70;80;89;90;100"
 
-# 8.7 is supported in the PyTorch main branch, but not 2.0.0
-
 ARG AOCL_BASE="/opt/aocl"
 ARG AOCL_VER="4.2.0"
 ARG AOCL_URL="https://download.amd.com/developer/eula/aocl/aocl-4-2/aocl-linux-aocc-4.2.0.tar.gz"
 
+# region Downloads
+
```
sangstar, line 21: What's this and `#endregion` for? Is that an IDE-specific thing?

torch/Dockerfile
```diff
@@ -223,11 +224,14 @@ RUN LLVM_VERSION='18' && \
 
 
 FROM builder-base-shared AS builder-base-arm64
-# There is currently no CPU BLAS used for ARM builds,
-# so this stage is just an alias
+ARG BUILD_TORCH_CUDA_ARCH_LIST
+ENV TORCH_CUDA_ARCH_LIST="9.0${BUILD_TORCH_CUDA_ARCH_LIST#* 9.0}"
+# There is currently no CPU BLAS used for ARM builds
```
sangstar, line 229: Does this change imply that this is no longer just an alias? I don't see `builder-base-arm64` being used elsewhere. What's the idea behind this?

sangstar (approved):
LGTM, few inquisitive questions.

torch/Dockerfile
Eta0: It is [used here as the base image for `builder-base`](https://github.com/coreweave/ml-containers/blob/69871cd8c14a14fbd17e59fd559760f1d1104107/torch/Dockerfile#L267) when the `${TARGETARCH}` is `arm64`. It is no longer just an alias because it now does something useful: it modifies the `${TORCH_CUDA_ARCH_LIST}` variable for `arm64` builds.

PR #98, opened 2025-06-17
feat(vllm-tensorizer): Upgrade vLLM version and Resolve Related Build Compatibility Issues

JustinPerlman:
This PR focuses on updating the vLLM version being used within the `vllm-tensorizer` Docker image and resolving the various bugs and compatibility issues that arose from this upgrade.

**Key Changes:**
  - **Updated Base Image:** Switched to `ghcr.io/coreweave/ml-containers/torch-extras:es-compute-12.0-67208ca-nccl-cuda12.9.0-ubuntu22.04-nccl2.27.3-1-torch2.7.1-vision0.22.1-audio2.7.1-abi1`. This moves from `torch:base` to `torch:nccl` for CUDA development tools and aligns with PyTorch 2.7.1 and CUDA 12.9.0.
  - **Upstream vLLM:** Updated vLLM source to pull from the official vllm-project/vllm.git repository (v0.9.1), rather than CoreWeave's vLLM fork
  - **Build Compatibility:** Resolved various compilation errors, including:
    - `FileNotFoundError: .../nvcc` and `CUDA::nvToolsExt not found` by ensuring correct `torch:nccl` base image.
    - `PyTorch version expected` errors by integrating `use_existing_torch.py` to handle vLLM's PyTorch version expectations.
    - Out-of-memory issues during Flash Attention compilation by setting `MAX_JOBS=16`.
    - Dockerfile Hygiene: Improved `Dockerfile` clarity by consistently uppercasing `AS` keywords.

### DIFF
--- a/.github/workflows/vllm-tensorizer.yml
+++ b/.github/workflows/vllm-tensorizer.yml
@@ -23,5 +23,6 @@ jobs:
       folder: vllm-tensorizer
       tag-suffix: ${{ matrix.vllm-commit }}
       build-args: |
-        VLLM_COMMIT_HASH=${{ matrix.vllm-commit }}
+        VLLM_COMMIT=${{ matrix.vllm-commit }}
+        FLASHINFER_COMMIT=${{ matrix.flashinfer-commit }}
         BASE_IMAGE=${{ matrix.base-image }}
--- a/.github/workflows/vllm-tensorizer.yml
+++ b/.github/workflows/vllm-tensorizer.yml
@@ -18,7 +18,7 @@ jobs:
     with:
       image-name: vllm-tensorizer
       folder: vllm-tensorizer
-      tag-suffix: ${{ inputs.commit || '85e2b7bb1380dd7096e0d6b64b0d7633b0a9db4a'}}
+      tag-suffix: ${{ inputs.commit || '5e4f640d22b0e4b50f506ce913d01fd471ad8086'}}
       build-args: |
-        COMMIT_HASH=${{ inputs.commit || '85e2b7bb1380dd7096e0d6b64b0d7633b0a9db4a'}}
+        COMMIT_HASH=${{ inputs.commit || '5e4f640d22b0e4b50f506ce913d01fd471ad8086'}}
         TRITON_COMMIT=96316ce5
--- a/.github/workflows/vllm-tensorizer.yml
+++ b/.github/workflows/vllm-tensorizer.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   get-config:
-    name: Get torch:base Config
+    name: Get vllm-tensorizer config
     uses: ./.github/workflows/read-configuration.yml
     with:
       path: ./.github/configurations/vllm-tensorizer.yml
--- a/.github/workflows/vllm-tensorizer.yml
+++ b/.github/workflows/vllm-tensorizer.yml
@@ -21,4 +21,3 @@ jobs:
       tag-suffix: ${{ inputs.commit || 'b6553be1bc75f046b00046a4ad7576364d03c835'}}
       build-args: |
         COMMIT_HASH=${{ inputs.commit || 'b6553be1bc75f046b00046a4ad7576364d03c835'}}
-        TRITON_COMMIT=96316ce5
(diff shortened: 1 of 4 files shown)


.github/workflows/vllm-tensorizer.yml
```diff
@@ -18,6 +18,7 @@ jobs:
     with:
       image-name: vllm-tensorizer
       folder: vllm-tensorizer
-      tag-suffix: ${{ inputs.commit || '19307ba71ddeb7e1cc6aec3c1baa8b50d59c1beb'}}
+      tag-suffix: ${{ inputs.commit || 'b6553be1bc75f046b00046a4ad7576364d03c835'}}
```
arsenetar, line 21: There shouldn't be a hardcoded commit here... this seems rather odd.

.github/workflows/vllm-tensorizer.yml
```diff
@@ -18,6 +18,7 @@ jobs:
...
       build-args: |
-        COMMIT_HASH=${{ inputs.commit || '19307ba71ddeb7e1cc6aec3c1baa8b50d59c1beb'}}
\ No newline at end of file
+        COMMIT_HASH=${{ inputs.commit || 'b6553be1bc75f046b00046a4ad7576364d03c835'}}
```
arsenetar, line 23: Same here, this ARG is also somewhat unclear as to what it is a commit hash of.

vllm-tensorizer/Dockerfile
```diff
@@ -1,66 +1,57 @@
-ARG BASE_IMAGE="ghcr.io/coreweave/ml-containers/torch-extras:es-22.04-58a49a2-base-cuda12.1.1-torch2.1.2-vision0.16.2-audio2.1.2-flash_attn2.4.2"
-
-FROM scratch as freezer
+ARG BASE_IMAGE="ghcr.io/coreweave/ml-containers/torch-extras:es-compute-12.0-67208ca-nccl-cuda12.9.0-ubuntu22.04-nccl2.27.3-1-torch2.7.1-vision0.22.1-audio2.7.1-abi1"
```
arsenetar, line 1: CI Should be passing base image as this will always be out of date.  This should move to something similar to what is done with the other torch images.  Also this looks like a non-main branch tag.  We need t o be building from main tags, not random branches.

arsenetar (commented):
How does the result of this build compare to the offical vllm image?  Do we have all the same libraries etc?  https://github.com/vllm-project/vllm/blob/main/docker/Dockerfile.

The CI here needs updates to build new version without manually specifying the version of base images to use.

I would also like to move to rename this to `vllm-openai` to align with the vllm project upstream image.

.github/workflows/vllm-tensorizer.yml
Eta0: It's a bit of a side-effect of not being able to declare variables very elegantly in GHA workflows. Normally this would go well in e.g. a build matrix declaration, but this currently only does one build rather than multiple.

It would also normally work better in a Dockerfile, but then the vLLM commit info wouldn't be accessible to the GHA workflow to add to the image tag.

arsenetar:
> What would you see as not manually specifying it? Automatically pulling the most recent image from `main`? Currently the only build in this repository that does that is `torch-extras` and it has some issues in practice, mainly that we often have changes in downstream images that require corresponding changes in the `torch` base image that will in turn require a new branch, so hardcoding it to look for images from `main` doesn't work very well for development

I would expect the CI to set the ARG for the base image, there are likely reasons where we will want to build a matrix of versions similarly to the other torch images.  Minimally that is something that should be changed to set that up for the CI to define/allow passing that.

With respect to upstream images breaking things and requiring updates, that is somewhat expected to happen. There is a difference between development and builds that should be considered validated released.  This repo's package structure does not really help delineate the differences.

Eta0:
> How does the result of this build compare to the offical vllm image? Do we have all the same libraries etc?

As far as I can tell, the only libraries we don't install that they do are `ffmpeg libsm6 libxext6 libgl1`, which is a bit of a strange set. I'm not sure we really need those, but if it comes up, we can add them. We also don't install `uv` to replace `pip`.

arsenetar:
> As far as I can tell, the only libraries we don't install that they do are ffmpeg libsm6 libxext6 libgl1, which is a bit of a strange set. I'm not sure we really need those, but if it comes up, we can add them. We also don't install uv to replace pip.

Out of all of that I could only think of ffmpeg as being of value if it's used by libraries for the multimodal.

PR #99, opened 2025-06-18
feat(torch): Support compute capability 12.0

Eta0:
# Compute Capability 12.0 + CUDA 12.9 Fixes

This change adds compute capability 12.0 (`sm_120`) to the list of compiled architectures for libraries in the `ml-containers/torch` and `ml-containers/torch-extras` images. It additionally fixes an issue with CUDA 12.9 support when compiling PyTorch extension modules by cherry-picking a patch from the `main` branch of the PyTorch repo into our `torch` build, and gets `torch` to use the system NVTX library when available (i.e. on CUDA 12.9).

`xformers` in `ml-containers/torch-extras` has also been updated to its v0.0.30 release, offering better compatibility with newer GPU architectures, PyTorch versions, and with vLLM.

Finally, this update drops compute capability 7.0 (e.g. V100), leaving 7.5 (e.g. Quadro RTX 4000/5000) as the lowest supported architecture.


torch/Dockerfile
```diff
@@ -53,6 +53,8 @@ RUN ./clone.sh pytorch/pytorch pytorch "${BUILD_TORCH_VERSION}" && \
     if [ "${BUILD_TORCH_VERSION}" = '2.5.1' ]; then \
       wget 'https://github.com/pytorch/pytorch/commit/1cdaf1d85f5e4b3f8952fd0737a1afeb16995d13.patch' -qO- \
       | git -C pytorch apply; \
+    elif [ "${BUILD_TORCH_VERSION}" = '2.7.1' ]; then \
+      git -C pytorch cherry-pick -n b74be524547f7f025066f19eda3b53a887c244ba; \
```
sangstar, line 57: What's going on here? Did this have to do with that build failure due to NVTX awhile back for that torch version?

torch/Dockerfile
```diff
@@ -10,8 +10,8 @@ ARG BUILD_FLASH_ATTN_VERSION="2.7.4.post1"
 ARG BUILD_FLASH_ATTN_3_VERSION="2.7.2.post1"
 ARG BUILD_TRITON_VERSION=""
 ARG BUILD_TRITON="1"
-ARG BUILD_TORCH_CUDA_ARCH_LIST="7.0 8.0 8.9 9.0 10.0+PTX"
-ARG BUILD_TRANSFORMERENGINE_CUDA_ARCH_LIST="70;80;89;90;100"
+ARG BUILD_TORCH_CUDA_ARCH_LIST="8.0 8.9 9.0 10.0 12.0+PTX"
+ARG BUILD_TRANSFORMERENGINE_CUDA_ARCH_LIST="80;89;90;100;120"
```
sangstar, line 14: No longer worth building for `7.0`?

sangstar (approved):
LGTM!

torch/Dockerfile
Eta0: It doesn't fit in the compiled torch binaries anymore without adding a compilation flag that could potentially slow down the other builds.

torch/Dockerfile
Eta0: Yes, it is cherry-picking a patch that avoids that failure. It's in PyTorch's `main` branch but not in any released version yet.

PR #102, opened 2025-07-21
feat: Update `torch` & `vllm-tensorizer` images, improve build process

Eta0:
# vLLM v0.9.2, TransformerEngine v2.4, Flash Attention 3 ~v2.7.4, xFormers v0.0.31, & build process updates

This change updates the `ml-containers/torch`, `ml-containers/torch-extras`, and `ml-containers/vllm-tensorizer` builds.

## `torch` & `torch-extras`

Previously, the version of xFormers in the `ml-containers/torch-extras` builds had issues with the version of Flash Attention 3 bundled in the `ml-containers/torch` builds. This has been resolved by updating both xFormers and Flash Attention 3 to newer versions. TransformerEngine has also been updated for compatibility with the newer Flash Attention 3 version.

### `torch-nightly`

Nightly builds (and future stable builds) of torch have been fixed. There was a change to the compilation steps of new versions of Triton ([triton-lang/triton](https://github.com/triton-lang/triton) PR 6381) required by the latest editions of PyTorch that had broken our existing build process. This was fixed, and our build process can now handle building old and new versions of Triton.

## `vllm-tensorizer`

vLLM can now use xFormers for supported models and code paths. It was previously unable to use xFormers because of its incompatibility with the bundled version of Flash Attention 3.

vLLM has also been updated to v0.9.2, up from v0.9.1, and the dependencies in the `vllm-tensorizer` image have been updated to match those in vLLM's own Dockerfile. Notably, we now include our own source build of `lmcache`, an optional dependency for vLLM with no ARM64 distribution. To facilitate building `lmcache` we also build its dependency `infinistore` from source, which similarly lacks an ARM64 distribution.

## Robustness

Builds of PyTorch and torch extensions such as Flash Attention and Flash Attention 3 now use a custom wrapper for their NVCC invocations capable of restarting the compilation of individual CUDA translation units under unusual failure conditions such as OOM crashes or compiler segmentation faults. The number of times to retry compilation is configurable, with an exponentially increasing wait time between attempts, up to one minute.

## Compute capabilities & multiarchitecture binaries

Where possible, packages built in the `ml-containers/torch` image now avoid compiling for the `sm_90`, `sm_100`, `sm_120`, `compute_90`, and `compute_100` architectures, preferring to compile for `sm_90a`, `sm_100a`, `sm_120a`, and `compute_120` instead. We already compiled for most `a`-suffixed architectures, but also redundantly compiled for their standard versions, which provide limited forward compatibility. That forward compatibility is not currently required in these cases, as we explicitly compile for later architectures as well.

Avoiding compiling for redundant architectures reduces issues with compilation time and binary size. Binary size per-package has become more of a concern with the increasing number of supported architectures, as the default (and most performant) [small code model](https://maskray.me/blog/2023-05-14-relocation-overflow-and-code-models) ceases to work for shared object files larger than ~2.1 GiB. That notwithstanding, the x86_64 build of Flash Attention 3 also now uses `-mcmodel=medium` to avoid any such linker issues, as that build is very large.

Filtering out compute capabilities at compilation time is achieved with the same NVCC wrapper script as used to implement retry logic—therefore, it even bypasses hardcoded architecture choices selected by extension authors. In the future, this could also be used for ARM64 builds as well, in order to more thoroughly avoid compiling for older architectures such as `sm_80`.


wbrown (approved):
Quite a few fixes here. Good job. :+1:

PR #103, opened 2025-07-25
build(vllm-tensorizer): Use `torch:nccl` for the final base image

Eta0:
# `vllm-tensorizer` NCCL base image

This change reverts the switch from `torch:nccl` to `torch:base` images from #101. Extra libraries are needed for vLLM from the `torch:nccl` images that are not in `torch:base`.


wbrown (approved):

arsenetar:
This seems to fix the issue it was intended to.

PR #104, opened 2025-07-29
feat(vllm-tensorizer): Update to vLLM v0.10.0 & `flashinfer` v0.2.8

Eta0:
# vLLM v0.10.0, `flashinfer` v0.2.8

This PR updates vLLM to its new version, along with updating its dependency `flashinfer` to match. vLLM v0.10.0 includes breaking changes, including ripping out parts of the vLLM V0 engine, so the builds from this PR branch should be given more scrutiny than other updates.

## Build changes

Updating `flashinfer` required changes to the compilation pipeline, as `flashinfer` v0.2.8 has issues with building for newer compute capabilities and on non-CUDA devices. In order to compile for the usual set of `ml-containers` compute capabilities (mainly the `a` variants of all modern compute capabilities), this PR copies over the `nvcc-wrapper.py` script from #102 to this container build's folder as well, and uses it to manipulate the NVCC `-gencode` flags present during the build. This would make more sense as a symlink, but Dockerfiles don't support following symlinks outside of their build context root directory.

### DIFF
--- a/.github/configurations/vllm-tensorizer.yml
+++ b/.github/configurations/vllm-tensorizer.yml
@@ -1,7 +1,7 @@
 vllm-commit:
-  - 'v0.9.2'
+  - 'v0.10.0'
 flashinfer-commit:
-  - 'v0.2.6.post1'
+  - 'v0.2.8'
 builder-base-image:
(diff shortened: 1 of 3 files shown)


wbrown (approved):
v0 is dead, long live v1!

PR #108, opened 2025-09-19
feat(vllm-tensorizer): Update to vLLM v0.10.2 and `flashinfer` v0.3.1

Eta0:
# vLLM v0.10.2

This change updates the `ml-containers/vllm-tensorizer` image to use vLLM v0.10.2. It also updates the bundled dependencies to match.

- A build of `NVIDIA/nixl` is now included in the container image at the path `/opt/nixl`
  - The corresponding Python bindings are at the nonstandard path `/opt/nixl/lib/python3/dist-packages/nixl`, but this directory has been added to the standard system Python search path so that `import nixl` works out of the box
- The optional dependencies `deepseek-ai/DeepEP` and `perplexityai/pplx-kernels` are not currently included as they rely on rather large dependencies themselves, such as `nvshmem`, but could be added if their functionality is needed

### DIFF
--- a/.github/configurations/vllm-tensorizer.yml
+++ b/.github/configurations/vllm-tensorizer.yml
@@ -1,7 +1,7 @@
 vllm-commit:
-  - 'v0.10.1'
+  - 'v0.10.2'
 flashinfer-commit:
-  - 'v0.2.11'
+  - 'v0.3.1'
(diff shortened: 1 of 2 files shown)


wbrown (approved):
Nice, looks like you got NIXL working!

PR #152, opened 2026-04-09
build(vllm-tensorizer): Integrate `sccache` & fix `flashinfer` build

Eta0:
# `sccache` Build Speedups & `flashinfer` Fix

This change builds off of the compiler caching work in #151 to bring fancy-pants object-storage-backed compiler caches to a `vllm-tensorizer` build near you. With this upgrade, builds will be better able to deduplicate work in compiling unchanged source files for CUDA, C++, and C code, even when the BuildKit builder layer cache gets invalidated, and in a way agnostic to the builder processing the job and whatever local cache mounts it may have.

Check out the PR description for #151 for a bit more information on the implementation. It transferred over pretty directly to this build, so there were no particularly novel tricks in the `vllm-tensorizer` integration. The biggest gotcha was that vLLM doesn't respect the `CUDACXX` environment variable to configure our `nvcc` wrapper, so `CMAKE_ARGS` had to be used instead.

Happy building!

## Bonus: `flashinfer` fix

The build logic for compiling `flashinfer` has been broken since 2026-03-24 on account of [`nvidia-nvshmem-cu12`](https://pypi.org/project/nvidia-nvshmem-cu12) and [`nvidia-nvshmem-cu13`](https://pypi.org/project/nvidia-nvshmem-cu13) putting out new v3.6.5 releases, which were incompatible with the `flashinfer` build, and broke compilation. Upstream, they worked around this by pinning those libraries to `<3.6` (see flashinfer-ai/flashinfer#2893), so this pins our installation of it similarly.

The reason this didn't cause all `vllm-tensorizer` builds since then to break is that that Dockerfile step has been hitting the build layer cache consistently enough since then that it's not actually *attempting* the `flashinfer` build most of the time, and is just relying on the last working result from it, but this won't last. At least [some of our builds since then](https://github.com/coreweave/ml-containers/actions/runs/23571486571/job/70300878330) have hit the error from this, though, which was probably rather cryptic. In any case, this gets it working again!


Rexwang8 (approved):
Great work!

PR #155, opened 2026-04-24
feat(sglang): Add sglang container build

JustinPerlman:
## What's included

  **Build pipeline**
  - `sglang/Dockerfile`, `build.bash`, `install.bash`
  - GitHub Actions workflow + `.github/configurations/sglang.yml` config (same pattern as vllm-tensorizer)
  - sccache S3 integration via the `common/` build context for compiler caching

  **Arch support**
  - Builds for both `linux/amd64` and `linux/arm64` (GH200/GB200)
  - ARM64 restricts CUDA arch list to `9.0 10.0+PTX` (no Ampere targets on ARM)
  - Builds xgrammar and decord from source on ARM64 (not available on PyPI for that arch)
  - xgrammar: uses `pip wheel --no-build-isolation --no-deps` to bypass exact nanobind version pin

  **Fixes**
  - Register PyTorch shared libs with ldconfig so torchao CUDA extensions load correctly at runtime
  - Use `sglang serve` entrypoint (replaces deprecated `python3 -m sglang.launch_server`)
  - Relax exact torch/torchao version pins in sglang's pyproject.toml to be compatible with the base image
  - `CMAKE_POLICY_VERSION_MINIMUM=3.5` to silence cmake 4.x breakage in FetchContent sub-projects

  **Performance**
  - `CMAKE_BUILD_PARALLEL_LEVEL` and `SGL_KERNEL_COMPILE_THREADS` tuned per arch
  - `--depth 1` shallow clone for xgrammar (tagged release)
  - Explicit parallelism for xgrammar and decord cmake builds


sglang/Dockerfile
```diff
@@ -1,24 +1,48 @@
-# syntax=docker/dockerfile:1.2
-ARG BASE_IMAGE
+# syntax=docker/dockerfile:1.10
+ARG BASE_IMAGE="ghcr.io/coreweave/ml-containers/torch-extras:7d4f07a-nccl-cuda12.9.1-ubuntu22.04-nccl2.29.7-1-torch2.11.0-vision0.26.0-audio2.11.0-abi1"
```
alexeldeib, line 2: shouldn't this come from the ci yaml? why change the default?

JustinPerlman: It does pull from that but you get a warning saying no default has been set if you don't also manually set it in the Dockerfile

sglang/build.bash
```diff
@@ -29,101 +24,63 @@ _PIP_INSTALL() {
...
-  cd triton
-  git checkout "${TRITON_COMMIT}"
-  _BUILD python |& _LOG triton.log
-); fi
-
-# flashinfer
-: "${FLASHINFER_COMMIT:?}"
-: "${CUTLASS_COMMIT:?}"
-(
-echo 'Building flashinfer-ai/flashinfer'
```
alexeldeib, line 48: what happened to these guys? no flashinfer/cutlass?

JustinPerlman: IIR, Flashinfer and cutlass no longer need to be built from source. 

- flashinfer_cubin-0.6.7.post2 and flashinfer_python-0.6.7.post2 get pulled in automatically as sglang dependencies during install.bash I believe
- sgl-kernel bundles the kernels that previously required building cutlass separately

alexeldeib: let's verify at runtime they do work/are in the image but okay

alexeldeib (approved):
seems reasonable, huzzah

sglang/Dockerfile
alexeldeib: that error is dumb imo, it is better to have it unset to avoid confusing defaults/mixed source of truth, but ack

JustinPerlman: Agreed. I wasn't sure which way to go so I just copied how its done for vLLM

PR #158, opened 2026-04-30
fix(vllm-tensorizer): preserve compute_80 for Marlin MoE

alexeldeib:
This fix follows directly from public vLLM source and the ml-containers NVCC wrapper semantics.

vLLM's Marlin MoE build requires an sm80 PTX fallback for non-FP8 WNA16 kernels:

- vLLM CMakeLists.txt sets MARLIN_MOE_ARCHS from "8.0+PTX".
- The same CMake block globs csrc/moe/marlin_moe_wna16/sm80_kernel_*.cu and applies MARLIN_MOE_ARCHS to those sources.
- vLLM's Marlin MoE generator emits sm80_kernel_* files for the non-FP8, non-sm75 cases.
- vLLM cmake/utils.cmake explicitly documents that a source arch of "8.0+PTX" remains selected for a newer target arch, so the intent is to compile an sm80 PTX fallback that newer GPUs can JIT.

ml-containers then wraps nvcc and filters gencode flags. The wrapper comments and implementation define NVCC_WRAPPER_FILTER_CODES as a denylist: values in that list are the gencode code= values to filter out, and transform_args drops matching -gencode/--generate-code entries.

The old arm64 filter listed both sm_80 and compute_80:

    sm_80;...;compute_80;...

That directly contradicted vLLM's build contract above by deleting the exact cubin/PTX fallback requested for Marlin MoE. The non-arm64 filter already omitted sm_80 and compute_80, so those builds already preserved the fallback.

The compiled extension confirms the effect. Inspecting Marlin MoE text sections in vllm/_moe_C.abi3.so from a build using the old arm64 denylist showed no sm80 Marlin MoE code:

    204 sm_120a
      6 sm_52

After removing sm_80/compute_80 from the denylist and rebuilding the same extension, the artifact contains the expected sm80 fallback again:

    204 sm_120a
    723 sm_80

The rebuild log also shows vLLM compiling the generated Marlin MoE sm80 sources that the old filter prevented from surviving as sm80/compute80 code, including:

    csrc/moe/marlin_moe_wna16/sm80_kernel_s8_u4b8_float16.cu.o
    csrc/moe/marlin_moe_wna16/sm80_kernel_s8_u4b8_bfloat16.cu.o
    csrc/moe/marlin_moe_wna16/sm80_kernel_bfloat16_u4b8_bfloat16.cu.o
    csrc/moe/marlin_moe_wna16/sm80_kernel_float16_u4b8_float16.cu.o

Remove sm_80/compute_80 from the arm64 denylist to align the platform filters and preserve vLLM's required Marlin MoE generated sm80 kernels.

### DIFF
--- a/vllm-tensorizer/Dockerfile
+++ b/vllm-tensorizer/Dockerfile
@@ -51,13 +51,16 @@ ENV TORCH_EXTENSION_SKIP_NVCC_GEN_DEPENDENCIES=1
 ARG TARGETPLATFORM
 # Switch 9.0, 10.0, and 12.0 to -a variants; preserve originals for PTX
 # Flashinfer v0.28.0 in particular can only build for 12.0a but not 12.0
+# vLLM Marlin MoE requires 8.0+PTX for non-FP8 WNA16 kernels. Keep the arm64
+# filter strict about unsupported sm80 SASS, but preserve compute80 PTX so
+# newer architectures can forward-JIT the generated sm80 kernels.
 RUN printf 'TORCH_CUDA_ARCH_LIST=' && \
     echo "${TORCH_CUDA_ARCH_LIST}" \
     | sed -E 's@\b(9|10|12)\.0\b@\1\.0a@g; s@\+PTX\b@@g' \
     | tee /opt/arch_list.txt && \
     printf 'NVCC_WRAPPER_FILTER_CODES=' && \
     if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
-      echo 'sm_80;sm_89;sm_90;sm_100;sm_120;compute_80;compute_89;compute_90;compute_100;compute_120'; \
+      echo 'sm_80;sm_89;sm_90;sm_100;sm_120;compute_89;compute_90;compute_100;compute_120'; \
     else \
       echo 'sm_90;sm_100;sm_120;compute_90;compute_100;compute_120'; \
     fi \


harubaru (dismissed):

vllm-tensorizer/Dockerfile
```diff
@@ -51,13 +51,17 @@ ENV TORCH_EXTENSION_SKIP_NVCC_GEN_DEPENDENCIES=1
...
+# filter aligned so newer architectures can forward-JIT the generated sm80
+# kernels.
 RUN printf 'TORCH_CUDA_ARCH_LIST=' && \
     echo "${TORCH_CUDA_ARCH_LIST}" \
     | sed -E 's@\b(9|10|12)\.0\b@\1\.0a@g; s@\+PTX\b@@g' \
     | tee /opt/arch_list.txt && \
     printf 'NVCC_WRAPPER_FILTER_CODES=' && \
     if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
-      echo 'sm_80;sm_89;sm_90;sm_100;sm_120;compute_80;compute_89;compute_90;compute_100;compute_120'; \
+      echo 'sm_89;sm_90;sm_100;sm_120;compute_89;compute_90;compute_100;compute_120'; \
```
alexeldeib: per @Eta0 can likely skip sm80 and keep compute_80, going to test that pre-merge since there is an existing workaround

alexeldeib: this is so confusing because it's a deny list but I am REMOVING compute_80* from the DENY list such that we ENABLE generating forward compatible PTX

I saw the diff and had to second guess myself for a few minutes there

Eta0 (approved):

PR #160, opened 2026-05-12
feat(vllm-tensorizer): Bump vLLM to v0.20.2 on CUDA 13.2 / Ubuntu 24.04

JustinPerlman:
## Summary

  - Bump vLLM to v0.20.2
  - Add a build matrix producing two variants, both on Ubuntu 24.04 / torch 2.11.0:
    - `v0.20.2-cuda13.2.1-ubuntu24.04`
    - `v0.20.2-cuda12.9.1-ubuntu24.04`

  ## Ubuntu 24.04 compatibility fixes

  - Remove `python3-pip` from apt in `builder-base` and add `rm -f /usr/lib/python3.*/EXTERNALLY-MANAGED` before pip bootstrap — on Ubuntu 24.04, apt-installed pip has no RECORD file and blocks pip self-upgrade
  - Purge `python3-jwt` in the final `base` stage before pip installs — same root cause: Debian-managed PyJWT has no RECORD file and blocks vLLM's dependency resolution
  - Fix `cuda-python` version spec from `~=${CUDA_VERSION}` to `~=${CUDA_VERSION%.*}` — patch-level CUDA versions (e.g. `13.2.1`) don't match available cuda-python releases; strip to major.minor
  - Install `wheel` package in `lmcache-builder` and restore it to `builder-base` pip install

Relevant information: https://github.com/vllm-project/vllm/pull/35386/changes/6c964bd0e4ed91594fa627a48200b22907856770

### DIFF
--- a/.github/configurations/vllm-tensorizer.yml
+++ b/.github/configurations/vllm-tensorizer.yml
@@ -5,6 +5,6 @@ flashinfer-commit:
 lmcache-commit:
   - 'v0.4.2'
 builder-base-image:
-  - 'ghcr.io/coreweave/ml-containers/torch:bc8c66e-nccl-cuda12.9.1-ubuntu22.04-nccl2.30.4-1-torch2.11.0-vision0.26.0-audio2.11.0-abi1'
+  - 'ghcr.io/coreweave/ml-containers/torch:bc8c66e-nccl-cuda13.2.1-ubuntu22.04-nccl2.30.4-1-torch2.11.0-vision0.26.0-audio2.11.0-abi1'
 final-base-image:
-  - 'ghcr.io/coreweave/ml-containers/torch:bc8c66e-nccl-cuda12.9.1-ubuntu22.04-nccl2.30.4-1-torch2.11.0-vision0.26.0-audio2.11.0-abi1'
+  - 'ghcr.io/coreweave/ml-containers/torch:bc8c66e-nccl-cuda13.2.1-ubuntu22.04-nccl2.30.4-1-torch2.11.0-vision0.26.0-audio2.11.0-abi1'
--- a/.github/configurations/vllm-tensorizer.yml
+++ b/.github/configurations/vllm-tensorizer.yml
@@ -1,10 +1,13 @@
-vllm-commit:
-  - 'v0.20.2'
-flashinfer-commit:
-  - 'v0.6.8'
-lmcache-commit:
-  - 'v0.4.2'
-builder-base-image:
-  - 'ghcr.io/coreweave/ml-containers/torch:bc8c66e-nccl-cuda13.2.1-ubuntu24.04-nccl2.30.4-1-torch2.11.0-vision0.26.0-audio2.11.0-abi1'
-final-base-image:
-  - 'ghcr.io/coreweave/ml-containers/torch:bc8c66e-nccl-cuda13.2.1-ubuntu24.04-nccl2.30.4-1-torch2.11.0-vision0.26.0-audio2.11.0-abi1'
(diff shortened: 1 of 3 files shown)


abatilo (commented):
This _seems_ reasonable to me but I'd feel better if @Eta0 could take a peek

JustinPerlman:
> This _seems_ reasonable to me but I'd feel better if @Eta0 could take a peek

Fair enough lol

alexeldeib:
Pure 13.2, no matrix with 12.9? 🫣I would really like having both options…if it’s a giant pain on vllm side it’s fine, but I think you then need to validate this actually works on b40/rtxp6000 with latest supported/installed drivers cw ships

alexeldeib:
I am still not aware of a cuda + driver combo that has decent support and works as expected, but haven’t followed too closely lately

alexeldeib (approved):
seems sane

.github/configurations/vllm-tensorizer.yml
Eta0, line 1: I'd personally suggest to not repeat yourself as much in this config file and to construct more parts of this dynamically, like the tag suffix, but not a hard requirement.

vllm-tensorizer/Dockerfile
```diff
@@ -22,9 +22,10 @@ RUN ldconfig
 
 RUN apt-get -qq update && \
     apt-get -qq install -y --no-install-recommends \
-      python3-pip git ninja-build cmake gcc-12 g++-12 && \
+      git ninja-build cmake gcc-12 g++-12 && \
     apt-get clean && \
-    pip3 install -U --no-cache-dir pip packaging 'setuptools>=77.0.3,<81.0.0' wheel setuptools_scm regex build
+    rm -f /usr/lib/python3.*/EXTERNALLY-MANAGED && \
+    python3 -m pip install -U --no-cache-dir pip packaging 'setuptools>=77.0.3,<81.0.0' wheel setuptools_scm regex build
```
Eta0, line 28: `rm -f /usr/lib/python3.*/EXTERNALLY-MANAGED ` and pip/setuptools installation and upgrading is already handled by the `torch` image, so you don't need to repeat those bits here.

vllm-tensorizer/Dockerfile
```diff
@@ -264,7 +265,11 @@ FROM ${FINAL_BASE_IMAGE} AS base
 
 WORKDIR /workspace
 
-RUN apt-get -qq update && apt-get install -y --no-install-recommends curl libsodium23 libnuma-dev && apt-get clean
+RUN apt-get -qq update && \
+    apt-get install -y --no-install-recommends curl libsodium23 libnuma-dev && \
+    apt-get purge -y python3-jwt && \
+    apt-get clean && \
+    rm -f /usr/lib/python3.*/EXTERNALLY-MANAGED
```
Eta0, line 272: Same comment as before: this `rm` is already handled by the base image.

vllm-tensorizer/Dockerfile
```diff
@@ -264,7 +265,11 @@ FROM ${FINAL_BASE_IMAGE} AS base
 
 WORKDIR /workspace
 
-RUN apt-get -qq update && apt-get install -y --no-install-recommends curl libsodium23 libnuma-dev && apt-get clean
+RUN apt-get -qq update && \
+    apt-get install -y --no-install-recommends curl libsodium23 libnuma-dev && \
+    apt-get purge -y python3-jwt && \
```
Eta0, line 270: What's that `apt-get purge -y python3-jwt` for? :eyes:

PR #162, opened 2026-05-17
feat(vllm): Bump llguidance for uri format

ritazh:
Needed by deepseek v4 customers

### DIFF
--- a/vllm-tensorizer/Dockerfile
+++ b/vllm-tensorizer/Dockerfile
@@ -326,6 +326,8 @@ ARG TARGETPLATFORM
 # * Gemma 4 needs transformers > 5.5.0
 # * Kimi K2.5 tool-call scramble fixed in transformers 5.5.4 (huggingface/transformers#45359)
 # * Kimi K2.5 registry-subprocess SIGSEGV on GB200 with nixl-cu12 1.0.1 (vllm-project/vllm#40642) — pin both to 1.0.0
+# * llguidance bumped to 1.7 — vLLM's pin (<1.4) lacks JSON Schema "format": "uri" support
+#   (added in llguidance v1.6.0; mirrors upstream vllm-project/vllm#42150)
 
 RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
       BITSANDBYTES_VER='0.42.0'; \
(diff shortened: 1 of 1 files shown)


alexeldeib:
Unrelated to your diff there’s a bug with build tags https://github.com/coreweave/ml-containers/blob/4ccd085c221a0d4743925896b8ac1d3b8605b036/.github/configurations/vllm-tensorizer.yml#L7

mind fixing that so we don’t dupe them

Happy to approve otherwise

@JustinPerlman fyi

ritazh:
@alexeldeib what should it be?

alexeldeib:
Should match the commit https://github.com/coreweave/ml-containers/blob/4ccd085c221a0d4743925896b8ac1d3b8605b036/.github/configurations/vllm-tensorizer.yml#L2 or be parsed out automatically

JustinPerlman:
> Should match the commit
> 
> https://github.com/coreweave/ml-containers/blob/4ccd085c221a0d4743925896b8ac1d3b8605b036/.github/configurations/vllm-tensorizer.yml#L2
> 
> or be parsed out automatically

Will fix this in my PR fyi @ritazh

.github/workflows/vllm-tensorizer.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     with:
       image-name: vllm-tensorizer
       folder: vllm-tensorizer
-      tag-suffix: ${{ matrix.tag-suffix }}
+      tag-suffix: ${{ format('{0}-cuda{1}-{2}', matrix.vllm-commit, matrix.cuda, matrix.os) }}
```
ritazh: @JustinPerlman PTAL

JustinPerlman: I'm going to implement it in a different way that derives the CUDA and OS from the base image itself instead of having to update the CUDA and OS manually in the `github/configurations/vllm-tensorizer.yaml` because manual updating that way is still subject to someone forgetting to update it.

See https://github.com/coreweave/ml-containers/pull/163

ritazh: ok then i will revert

ritazh:
@alexeldeib @JustinPerlman this one should be good to merge now

alexeldeib (approved):

PR #163, opened 2026-05-18
feat(vllm-tensorizer): Add DeepEP + audio, fix tag drift, drop redundant cleanup

JustinPerlman:
Four improvements to the `vllm-tensorizer` image:

  - **DeepEP + NVSHMEM** for EP all2all on InfiniBand. Ports @kptkin's [#149](https://github.com/coreweave/ml-containers/pull/149), which had stalled unmerged. Enables `--all2all-backend deepep_high_throughput` for
   MoE expert-parallel workloads (DeepSeek + EP/EPLB).
  - **vLLM audio extra** installed alongside `tensorizer` — fixes runtime errors like `Please install vllm[audio] for audio support` (pulls `av`, `scipy`, `soundfile`, `mistral_common[audio]`). https://coreweave.slack.com/archives/D0A8XNCQ0M6/p1778868874150649
  - **Tag-suffix is now derived from the base image** to fix the drift @alexeldeib called out in [#162](https://github.com/coreweave/ml-containers/pull/162#issuecomment-4475194630). Previously `tag-suffix` was
  hand-written and went stale when only `vllm-commit` got bumped — the v0.21.0 builds on main are tagged `v0.20.2-...`. Now the suffix is `${vllm-commit}-cuda<X.Y.Z>-ubuntu<X.YY>`, regex-extracted from
  `builder-base-image` via a yq filter passed to `read-configuration`. Bumping the base image automatically rotates the tag; no manual sync.
  - **Drops the redundant `rm -f /usr/lib/python3.*/EXTERNALLY-MANAGED` lines** in `builder-base` and final `base`. The torch base image at `bc8c66e` already removes that file in both its `builder-base-shared` and
  final stages, and nothing in our apt operations restores it. Followup on @Eta0's review feedback from [#160](https://github.com/coreweave/ml-containers/pull/160).


vllm-tensorizer/Dockerfile
```diff
@@ -268,15 +284,14 @@ WORKDIR /workspace
 RUN apt-get -qq update && \
     apt-get install -y --no-install-recommends curl libsodium23 libnuma-dev && \
     apt-get purge -y python3-jwt && \
-    apt-get clean && \
```
alexeldeib, line 271: What’s up here dropping externally managed

JustinPerlman: Regarding `apt-get purge -y python3-jwt`

With Ubuntu 24 base images I was prev getting `Cannot uninstall PyJWT 2.7.0. The package's contents are unknown: no RECORD file was found for PyJWT.`

Purging the Debian package removed it cleanly, letting pip install its own without the error.

alexeldeib (approved):

PR #165, opened 2026-05-21
feat(sglang): Bump to v0.5.12 on CUDA 13.2.1 base

JustinPerlman:
## Summary

  Bump the `sglang` image to **sglang v0.5.12** on a **CUDA 13.2.1 / Ubuntu 24.04 / NCCL 2.30.4 / Torch 2.11.0** base, and modernize the build to track upstream's v0.5.12 dependency surface.

  ## Notable changes

  **Bumps**
  - `sglang`: v0.5.10.post1 → v0.5.12 (commit `127b9e3`)
  - Base image: `torch-extras:7d4f07a-nccl-cuda12.9.1-ubuntu22.04-…` → `torch-extras:bc8c66e-nccl-cuda13.2.1-ubuntu24.04-nccl2.30.4-1-torch2.11.0-…`
  - Output tag suffix: `v0.5.10.post1-cuda12.9.1-torch2.11.0` → `v0.5.12-cuda13.2.1-torch2.11.0`

  **New build deps for v0.5.12**
  - v0.5.12 ships a Rust gRPC extension (`rust/sglang-grpc`, edition 2024, channel pinned to 1.90) built via setuptools-rust. `build.bash` now installs:
    - `setuptools-rust>=1.10` (host env, required because we build with `--no-isolation`)
    - `protobuf-compiler` via apt (needed by `tonic-build`/`prost-build`)
    - Rust 1.90 via rustup
  - Pip pins migrate to follow upstream exactly:
    - `xgrammar`: 0.1.32 → 0.2.0
    - `torchao`: 0.9.0 → 0.17.0
    - `transformers`: pinned to 5.6.0
    - `cuda-python`: ≥13.0 (satisfied by the CUDA 13 base image)

  **Removed: from-source ARM64 builds**
  - v0.5.12 publishes `xgrammar` aarch64 wheels on PyPI; replaces the in-tree build.
  - Upstream switched ARM64 from `decord` to `decord2` (PyPI aarch64 wheels) and switched x86_64 to `torchcodec`. The from-source `decord` build, the `DECORD_COMMIT` build-arg, and the decord-specific apt deps in
  `install.bash` are gone.
  - `install.bash` no longer carries an explicit pip-install override list; sglang's pyproject pins now cover `xgrammar`/`llguidance`/`torchao`/`outlines`/etc. directly. Torch-extension `ldconfig` fix kept.
  - The pin-relaxing seds in `build.bash` are removed — the base image now matches upstream's torch-family pins.

  ## Test

  - Tested by direct Helm install using image `ghcr.io/coreweave/ml-containers/sglang:jperlman-sglang-v0.5.12-4ba3647-v0.5.12-cuda13.2.1-torch2.11.0`. Deployed `microsoft/Phi-4-mini-instruct` (1×H100, `engine: sglang`) and verified `/v1/chat/completions` returns coherent responses.

### DIFF
--- a/sglang/build.bash
+++ b/sglang/build.bash
@@ -27,7 +27,7 @@ _PIP_INSTALL() {
 # Python build deps. `setuptools-rust>=1.10` is required for sglang's gRPC
 # Rust extension (rust/sglang-grpc) since v0.5.12; we build with `--no-isolation`
 # so it must be present in the host environment.
-_PIP_INSTALL -U pip setuptools wheel build ninja \
+_PIP_INSTALL -U pip 'setuptools<82' wheel build ninja \
   'scikit-build-core>=0.10' 'setuptools-scm>=8.0' 'setuptools-rust>=1.10'
 
 # protobuf-compiler: needed by tonic-build (via prost-build) when compiling the
@@ -55,6 +55,10 @@ sed -Ei \
   -e 's@"torchao==[0-9]+\.[0-9]+\.[0-9]+"@"torchao>=0.17.0"@' \
   -e 's@"torchcodec==[0-9]+\.[0-9]+\.[0-9]+@"torchcodec@' \
   python/pyproject.toml
+
+# Surface the Rust toolchain file at the repo root so rustup's CWD-upward
+# walk finds it when setuptools-rust invokes cargo from python/.
+ln -sf rust/sglang-grpc/rust-toolchain.toml .
 
 # Build sgl-kernel (scikit-build-core + CMake; deps via FetchContent).
 (
--- a/sglang/build.bash
+++ b/sglang/build.bash
@@ -24,10 +24,27 @@ _PIP_INSTALL() {
   "$@"
 }
 
-_PIP_INSTALL -U pip setuptools wheel build ninja 'scikit-build-core>=0.10' 'setuptools-scm>=8.0'
+# Python build deps. `setuptools-rust>=1.10` is required for sglang's gRPC
+# Rust extension (rust/sglang-grpc) since v0.5.12; we build with `--no-isolation`
+# so it must be present in the host environment.
+_PIP_INSTALL -U pip setuptools wheel build ninja \
+  'scikit-build-core>=0.10' 'setuptools-scm>=8.0' 'setuptools-rust>=1.10'
 # Do not install cmake via pip — the base image provides cmake from the Kitware
 # PPA (3.x). pip's cmake package installs 4.x which removed backward compat
 # with cmake_minimum_required(VERSION < 3.5) used by some sub-project deps.
+
+# protobuf-compiler: needed by tonic-build (via prost-build) when compiling the
+# sglang-grpc Rust crate.
+apt-get -qq update && apt-get -q install --no-install-recommends -y \
+  protobuf-compiler
+
+# Rust toolchain: sglang/rust/sglang-grpc requires edition 2024 (rustc >= 1.85);
+# its rust-toolchain.toml pins channel 1.90, which rustup will fetch lazily.
+curl --proto '=https' --tlsv1.2 --retry 3 --retry-delay 2 -sSf https://sh.rustup.rs \
+  | sh -s -- -y --no-modify-path --profile minimal --default-toolchain 1.90
+export PATH="/root/.cargo/bin:${PATH}"
+rustc --version
+cargo --version
 
 # sglang (includes sgl-kernel)
 : "${SGLANG_COMMIT:?}"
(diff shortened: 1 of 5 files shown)


sglang/build.bash
```diff
@@ -37,15 +54,9 @@ git clone --recursive --filter=blob:none https://github.com/sgl-project/sglang
 cd sglang
 git checkout "${SGLANG_COMMIT}"
 
-# Relax exact torch-family version pins to be compatible with the base image
-sed -Ei \
-  -e 's@"torch==[0-9]+\.[0-9]+\.[0-9]+"@"torch>=2.8.0"@' \
-  -e 's@"torchaudio==[0-9]+\.[0-9]+\.[0-9]+"@"torchaudio>=2.8.0"@' \
-  -e 's@"torchao==[0-9]+\.[0-9]+\.[0-9]+"@"torchao>=0.9.0"@' \
-  -e 's@"torchcodec==[0-9]+\.[0-9]+\.[0-9]+@"torchcodec@' \
-  python/pyproject.toml
```
Eta0, line 46: Why is this getting deleted? This will be necessary as soon as it gets out of sync again. Just update the version numbers specified.

sglang/build.bash
```diff
@@ -37,15 +54,9 @@ git clone --recursive --filter=blob:none https://github.com/sgl-project/sglang
...
-
-# Build sgl-kernel (scikit-build-core + CMake; deps via FetchContent)
+# Build sgl-kernel (scikit-build-core + CMake; deps via FetchContent).
+# Published wheel name is `sglang-kernel` (sglang v0.5.12+); pyproject pins
+# sglang-kernel==0.4.2.post2, satisfied by this in-tree build.
```
Eta0: This comment is strange. It'll be outdated as soon as there's a version bump, and don't people assume by default that they're reading an internally-consistent build? Is this supposed to warn somebody not to do something in particular? If so, what?

sglang/install.bash
```diff
@@ -11,23 +11,10 @@ _PIP_INSTALL() {
   "$@"
 }
 
+# pip resolves sglang's runtime deps (xgrammar, llguidance, torchao,
+# transformers, decord2/torchcodec, flashinfer, …) from PyPI using the pins
+# in upstream sglang's pyproject.toml — no explicit overrides needed here.
```
Eta0: Why leave a comment about what we're not doing?

sglang/build.bash
```diff
@@ -24,11 +24,28 @@ _PIP_INSTALL() {
...
+apt-get -qq update && apt-get -q install --no-install-recommends -y \
+  protobuf-compiler
+
+# Rust toolchain: sglang/rust/sglang-grpc requires edition 2024 (rustc >= 1.85);
+# its rust-toolchain.toml pins channel 1.90, which rustup will fetch lazily.
+curl --proto '=https' --tlsv1.2 --retry 3 --retry-delay 2 -sSf https://sh.rustup.rs \
+  | sh -s -- -y --no-modify-path --profile minimal --default-toolchain 1.90
+export PATH="/root/.cargo/bin:${PATH}"
+rustc --version
+cargo --version
```
Eta0: The comment here contradicts the code block immediately below it. It says that `rustup` will fetch the toolchain lazily, but you immediately invoke `rustc` after it, showing that it was installed eagerly. Presumably that's since you include `--default-toolchain 1.90`, but we have no business pinning a rust toolchain version if `--default-toolchain none` would work and fetch the correct one at build time anyway.

Eta0 (changes requested):

sglang/build.bash
Eta0: Okay, given that after relaxing this, your latest build failed with the following:

```
running build_ext
running build_rust
error: rustup could not choose a version of rustc to run, because one wasn't specified explicitly, and no default is configured.
help: run 'rustup default stable' to download the latest stable release of Rust and set it as your default toolchain.
error: can't find Rust compiler
```

I'd double-check this original claim that it pins the version at 1.90 itself. If it doesn't actually do that, then we can probably just run with its default-default (probably the latest `rustc` available). If it's trying to pin one and failing, we should probably at most just go parse the TOML file at build-time or something to grab the version so that we don't need to update a hardcoded value here each time.

JustinPerlman: Oops just saw this. 

https://github.com/sgl-project/sglang/blob/main/rust/sglang-grpc/rust-toolchain.toml does pin 1.90

I did something kinda similar https://github.com/coreweave/ml-containers/pull/165/changes#diff-4a74cdbbebd0f1c7f5cbb0785be6ff9aa628859c6d6f472642ca0cb5fbbe09deR59-R61 👀

Eta0 (approved):

sglang/build.bash
Eta0: Sounds good if it works, then!

PR #167, opened 2026-05-27
build(vllm-tensorizer): Add support for `sm_103a`

Eta0:
# vLLM `sm_103a` support

This change adds a `sed` hack to the vLLM build to prevent it from filtering CC 10.3 out of its list of compute capabilities that it is being compiled for. This will allow vLLM to better support B300 GPUs.


shajrawi (approved):
Thanks Eta!

alexeldeib (approved):
yay

Eta0:
Builds finished successfully, and I confirmed that all the main vLLM shared object files now contain `sm_103a` SASS in addition to all the arches they had previously. Should be all good.

PR #171, opened 2026-06-03
build(vllm-tensorizer): Update lmcache version for cu13. v0.4.6 uses cupy-cuda13x

gshtrasberg:


.github/configurations/vllm-tensorizer.yml
kptkin: do you need to update this one as well?

gshtrasberg: No, this one targets cuda 12, and lmcache 0.4.6 hard codes cu13

gshtrasberg: Actually no, I think you're right, 0.4.6 will detect during build time

alexeldeib (approved):
seems sane

PR #173, opened 2026-06-11
build: Install NIXL against HPC-X mt

gshtrasberg:
nccl-tests installs a custom HPC-X system-wide (base for torch -- base for vllm-tensorizer). The prebuilt nixl PyPI wheel (pulled in transitively by LMCache via nixl>=1.1.0) bundles its own UCX, which is incompatible with this system HPC-X — 
GPU memory registration fails:
```
from nixl._api import nixl_agent, nixl_agent_config; import torch
a = nixl_agent("p", nixl_agent_config(backends=["UCX"])); t = torch.zeros(1024, device="cuda")
print("OK" if a.register_memory([(t.data_ptr(), t.numel()*t.element_size(),0,"")],"VRAM") else "FAILED")
```
Fix: build nixl from source against the system HPC-X multi-threaded UCX (/opt/hpcx/ucx/mt; the default /opt/hpcx/ucx is single-threaded, which NIXL needs MT over), don't bundle UCX in the wheel, and prepend /opt/hpcx/ucx/mt/lib to LD_LIBRARY_PATH so the MT libucp is loaded at runtime. The source build replaces the prebuilt wheel after LMCache is installed.

For the previous
> * Kimi K2.5 registry-subprocess SIGSEGV on GB200 with nixl-cu12 1.0.1 (vllm-project/vllm#40642) — pin both to 1.0.0

Verified that Kimi-K2.5 starts fine on 4xGB200 with this image

### DIFF
--- a/vllm-tensorizer/Dockerfile
+++ b/vllm-tensorizer/Dockerfile
@@ -277,39 +277,35 @@ RUN --mount=type=bind,from=vllm-downloader,source=/git/vllm/tools/ep_kernels,tar
     mkdir -p /nvshmem_libs && \
     cp /tmp/deepep_build/nvshmem/lib/libnvshmem*.so* /nvshmem_libs/
 
+# Build a NIXL wheel linked against HPC-X's multi-threaded UCX (/opt/hpcx/ucx/mt). The PyPI
+# nixl-cuNN wheels bundle their own UCX, which clashes with this image's HPC-X UCX and can't
+# register GPU memory. UCX+POSIX are the only plugins needed; CUDA arch stays at NIXL's default
+# (sm_80+) to keep the wheel multi-arch.
 FROM builder-base AS nixl-builder
 RUN apt-get -qq update && \
     apt-get -q install --no-install-recommends --no-upgrade -y \
-      build-essential pkg-config && \
+      build-essential pkg-config git && \
     apt-get clean && \
-    python3 -m pip install --no-cache-dir meson ninja pybind11 cmake
+    python3 -m pip install --no-cache-dir meson meson-python ninja pybind11 tomlkit patchelf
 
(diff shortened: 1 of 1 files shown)


alexeldeib (approved):
bit gnarly but yay

vllm-tensorizer/Dockerfile
```diff
@@ -387,4 +381,27 @@ RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
...
+RUN --mount=type=bind,from=nixl-builder,source=/wheels,target=/tmp/wheels \
+    python3 -m pip uninstall -y nixl nixl-cu12 nixl-cu13 2>/dev/null; \
+    rm -rf /usr/local/lib/python3.12/dist-packages/nixl* /usr/local/lib/python3.12/dist-packages/.nixl* && \
+    python3 -m pip install --no-cache-dir --no-deps /tmp/wheels/*.whl && \
+    python3 -m pip install --no-cache-dir --no-deps nixl && \
+    # Drop the meta's nixl-cuNN deps so downstream installs (e.g. modelexpress's `nixl[cu12]`) can't
+    # pull a prebuilt bundled-UCX wheel back in; our build is the only backend.
+    sed -i '/^Requires-Dist: nixl-cu1[23]/d' /usr/local/lib/python3.12/dist-packages/nixl-*.dist-info/METADATA && \
+    # Assert the UCX plugin links HPC-X's MT UCX (LD_LIBRARY_PATH below selects it at runtime).
+    P="$(find /usr/local/lib/python3.12/dist-packages -name 'libplugin_UCX.so' | head -1)" && \
```
alexeldeib, line 396: nit: hardcoding python version, think this is the only place we do it

PR #177, opened 2026-06-29
feat(vllm): Upgrade to vLLM v0.24.0

JustinPerlman:


alexeldeib (approved):
noice, looking forward to dropping some patches (and then less looking forward to rebasing some)

PR #178, opened 2026-06-30
feat(vllm): Use exact commit for vLLM

JustinPerlman:
Uses the exact upstream commit for vLLM build. Also update to newest version of v0.24.0 (prerelease v0.24.0 got added on to).


alexeldeib (approved):
seems okay, annoying they modified it

PR #180, opened 2026-07-06
feat(vllm-tensorizer): Build dynamo-vllm as a multi-stage target

JustinPerlman:
## What

Fold the `dynamo-vllm` image into the vLLM `Dockerfile` as a multi-stage build so that both `vllm-tensorizer` and `dynamo-vllm` are built from the same file and stay version-locked.

Resolves [INF-748](https://coreweave.atlassian.net/browse/INF-748).

## Why

Building vLLM and Dynamo from one Dockerfile keeps their versions in lockstep so they can be rolled back as a unit. (Rationale from [coreweave/infr#800](https://github.com/coreweave/infr/pull/800#issuecomment-4838641336).)

## Changes

- **`vllm-tensorizer/Dockerfile`** — Add a `dynamo` stage (`FROM base AS dynamo`) that layers `ai-dynamo` + deps on top of the vLLM `base` stage, mirroring the standalone infr Dockerfile (same packages, workdir, labels). The build-time smoke test additionally re-verifies `import nixl`, since Dynamo's deps are installed unconstrained on top of `base`'s carefully-pinned runtime — this fails the build fast if Dynamo ever perturbs the HPC-X MT-UCX nixl setup.
- **`.github/workflows/build.yml`** — Add two optional inputs to the reusable build workflow:
  - `target` — passed as `--target` to `docker/build-push-action`.
  - `extra-cache-from-image-name` — adds a second read-only `cache-from` registry ref so a derived image can reuse a parent image's compiled layers.
- **`.github/workflows/vllm-tensorizer.yml`** — Pin the existing build to `target: base` (preserving the current `vllm-tensorizer` image exactly), and add a `build-dynamo` job that builds `target: dynamo` as image `dynamo-vllm`. It `needs: [get-config, build]` and sets `extra-cache-from-image-name: vllm-tensorizer`, so it reuses the just-built vLLM cache and only builds the thin Dynamo layer instead of recompiling vLLM.
- **`.github/configurations/vllm-tensorizer.yml`** — Add `dynamo-version` to both CUDA entries (12.9 + 13.2) so the Dynamo version lives next to `vllm-commit` and updates together.

[INF-748]: https://coreweave.atlassian.net/browse/INF-748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/vllm-tensorizer.yml
```diff
@@ -44,3 +45,31 @@ jobs:
...
+  build-dynamo:
+    uses: ./.github/workflows/build.yml
+    needs: [get-config, build]
+    strategy:
+      matrix: ${{ fromJSON(needs.get-config.outputs.config) }}
+    secrets: inherit
+    with:
+      image-name: dynamo-vllm
+      folder: vllm-tensorizer
+      target: dynamo
```
alexeldeib, line 62: why are they even different? can't you use one for everything, once you add the dynamo deps on top? or is there some other issue?

alexeldeib: the final dockerfile sure looks like you could just build 1 that is compatible with both

JustinPerlman: One image with vllm+dynamo for everything (and not have a non-dynamo vllm build)?

ritazh:
Thanks for the PR @JustinPerlman 

dynamo-vllm should include the vllm patches/plugins/etc we maintain in https://github.com/coreweave/infr/tree/main/images/vllm This is why currently docker.artifacts.coreweave.com/cw-infr-oci/dynamo-vllm is built from docker.artifacts.coreweave.com/cw-infr-oci/vllm as you see in https://github.com/coreweave/infr/blob/main/images/dynamo-vllm/Dockerfile#L2

PR #186, opened 2026-07-17
chore(renovate): semi-automate inference engine updates

alexeldeib:
Automate the first step of evaluating new vLLM and SGLang releases while preserving human approval for anything that is not a simple bump.

- Scope Renovate to the two inference configuration files and their two upstream GitHub release feeds.
- Resolve release tags to immutable source SHAs and update both values together.
- Apply the shared seven-day release-age policy, build candidate branches before requesting review, and never automerge.
- Run hourly at minute 17; no-update runs do not trigger image builds.
- Default manual dispatches to dry-run and unattended runs to concise `info` logging.
- Add the existing inference team as CODEOWNERS for SGLang paths.

Validation:

- [Renovate full dry run](https://github.com/coreweave/ml-containers/actions/runs/29853956406): authenticated as `dependa-jr[bot]`, loaded the shared preset, proposed SGLang `v0.5.15.post1` plus commit `0b3bb0cbe31873994c9f989fddfe2f87ca839fdd`, left vLLM at `v0.25.1`, and performed no writes.
- [SGLang build](https://github.com/coreweave/ml-containers/actions/runs/29855149629): passed on the final commit.
- [vLLM-tensorizer build](https://github.com/coreweave/ml-containers/actions/runs/29855149625): passed on the final commit.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,7 @@
 # Inference specific
+/sglang/ @coreweave/inference
+/.github/configurations/sglang.yml @coreweave/inference
+/.github/workflows/sglang.yml @coreweave/inference
 /vllm-tensorizer/ @coreweave/inference
 /.github/configurations/vllm-tensorizer.yml @coreweave/inference
 /.github/workflows/vllm-tensorizer.yml @coreweave/inference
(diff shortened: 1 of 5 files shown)


alexeldeib:
dunno if there's a reason we don't use renovate otherwise besides the obvious complexity, but this seems like an ez quick win even if it fails every time, since we all do the same thing as step 1

I saw #156 which was bulk-PR'd and does not add an explicit workflow, which I did, so I assume this isn't like, insane

JustinPerlman (approved):

PR #199, opened 2026-07-29
fix(vllm): Pipe CUDA_VERSION into the FlashInfer local-version sed

abatilo:
The substitution read empty stdin, so FLASHINFER_LOCAL_VERSION was always
blank and the version suffix was silently dropped. That made the wheels built
here indistinguishable from the ones on PyPI, so pip treated vLLM's pinned
flashinfer-python as already satisfied and the image ran upstream FlashInfer
rather than its own build -- including any patch applied to it.

The wheels are now 0.6.15.post1+cu132, which still satisfies vLLM's pin and
sorts higher, so ours install without --force-reinstall. The version string
also makes it obvious from pip list which build is running.


abatilo:

- #199 ◀
    - #200
        - #201

nehashah-eng (dismissed):

alexeldeib (approved):
yeehaw mode

PR #200, opened 2026-07-29
build(vllm): Build only Kimi-K3 and drop the upstreamed FlashInfer patch

abatilo:
Kimi-K3 only exists on an unmerged vLLM branch, so both released variants are
replaced by one entry pinned to a commit on that branch.

Renovate does not track that entry at all. Any bump would need the SiTU patch
re-ported by hand first, so an automated PR could only ever be noise. The entry
simply carries no renovate-release marker, which is sufficient because the
custom regex manager is the only manager enabled for this repo. Nothing here
watches vLLM releases now, so check for a K3-bearing release by hand.

CUDA 13 only, not by preference: K3's kernels require it, so a CUDA 12 image
builds and then fails at model init. Nothing consumed the CUDA 12 build.

The flashinfer-v0.6.14-deepseekv3-no-group-routing patch is dropped because
the NaN fix it carried landed upstream before v0.6.15.post1. Upstream did not
take the patch's other half, so re-adding a DeepSeek-V3 entry means re-porting
it.


abatilo:

- #199
    - #200 ◀
        - #201

PR #201, opened 2026-07-29
feat(vllm): Reconstruct FlashInfer SiTU support for Kimi-K3

abatilo:
Kimi-K3's MoE activation is SiTU, and public FlashInfer cannot serve it: no
published artifact contains the kernels. flashinfer#4180 is the public attempt
and is unbuildable for the same reason, so it is not a substitute for the
patch added here.

Instead this reconstructs the dual-mode loader that ships in vLLM's own
vllm/vllm-openai:kimi-k3 images -- a patch adding the SiTU activation and a
private JIT mode, the matching cubin pool copied from that image by digest,
and a DeepGEMM repin to the fork rev vLLM's own build already vendors. The
pool is an overlay, so the public artifact pin stays as it is. vLLM's kimi-k3
branch requires this API unconditionally; it is not optional.

Verified serving Kimi-K3 on 2x GB300 with flashinfer_trtllm at TP8 and DSpark
speculation. The private module JIT-compiles on the first request, a few
minutes per node; prewarming it is follow-up.

Once a release carries K3 and FlashInfer ships SiTU publicly, remove this
patch, the pool stage, FLASHINFER_PRIVATE_CUBIN_DIR, and the DeepGEMM pin. If
the release still requires the private API, re-key the patch rather than drop
it.

### DIFF
--- a/vllm-tensorizer/Dockerfile
+++ b/vllm-tensorizer/Dockerfile
@@ -98,6 +98,11 @@ RUN git clone --filter=tree:0 --no-single-branch --no-checkout \
     git submodule update --init --recursive --jobs 8 \
       --depth 1 --filter=tree:0
 
+# Kimi-K3's MoE activation is SiTU which is not in any official release
+COPY flashinfer-v0.6.15.post1-situ-private-cubin-pool.patch /git/patch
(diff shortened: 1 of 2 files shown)


abatilo:

- #199
    - #200
        - #201 ◀

PR #202, opened 2026-07-30
feat(vllm-tensorizer): Bump DeepGEMM to a6b593d2 for SM12x attention kernels

surajssd:
Align `DEEPGEMM_COMMIT` with upstream vLLM main's pin (`cmake/external_projects/deepgemm.cmake`). The previous pin `f5a76426` ships paged-MQA logits kernels for SM90/SM100 only; `a6b593d2` (89 commits ahead, on the DeepGEMM nv_dev line — also reachable in vllm-project/DeepGEMM via nv_dev+situ) adds the arch-generalized smxx dispatch layer and the SM120 variants (`sm120_fp8_paged_mqa_logits`, `sm120_fp4_paged_mqa_logits`).

Without them, DeepSeek-V4's sparse-attention Lightning indexer asserts 'Unsupported architecture' (csrc/apis/attention.hpp:219) on RTX PRO 6000 (B40 / SM120), blocking DeepSeek-V4-family NVFP4 checkpoints on that hardware.

`a6b593d2` is upstream vLLM main's exact pin, so API compatibility with recent vLLM is by construction.

PR #203, opened 2026-07-30
feat(vllm): Serve Kimi-K3 from public upstream SiTU

abatilo:
Kimi-K3 landed in vLLM main ([#50000](https://github.com/vllm-project/vllm/pull/50000)) and its SiTU MoE kernels in FlashInfer main ([#4180](https://github.com/flashinfer-ai/flashinfer/pull/4180)), and NVIDIA's public trtllm-gen artifact now ships `siTuGlu` cubins. Every reason this image carried a private SiTU reconstruction is gone, so the reconstruction goes with it.

**Deleted (1,762 lines):**
- the 1,736-line `flashinfer-v0.6.15.post1-situ-private-cubin-pool.patch`
- the `vllm/vllm-openai:kimi-k3` stage it pulled a private cubin pool from
- `FLASHINFER_PRIVATE_CUBIN_DIR` and its `COPY`

vLLM no longer passes `is_private` at all.

**Pins are main commits, not tags.** Neither project has tagged these merges, and both cut release tags off release branches, so a main merge never lands in an existing tag. FlashInfer has to stay ahead of the version vLLM's own requirements pin (which predates SiTU); our wheels install after vLLM's and win on version, so no pin relaxation is needed. The `TODO(kimi-k3)` to restore `github-releases` tracking stands.

torch 2.13.0 / vision 0.28.0 are unchanged — vLLM main still requires exactly those, so the existing base image already matches.

**Moving to vLLM main also moved its EP-kernel toolchain, which needed two fixes:**
- The DeepEP and NVSHMEM overrides are **dropped rather than bumped**. vLLM's `tools/ep_kernels` script patches DeepEP's sources, so it must pick a matching rev; our stale `73b6ea4` (2025-10) predated the `symmetric.hpp` the newer script patches. `NVSHMEM_VER` merely restated the script's own default. The script owns both from here, which is what stops this recurring on the next vLLM bump.
- The newer DeepEP resolves NCCL from a pip wheel, which this image does not have. The torch-nccl base compiles torch with `USE_SYSTEM_NCCL` against the fabric-tuned NCCL 2.30.4-1 from CoreWeave's nccl-tests image, so there is exactly one NCCL and its version is pinned in the base tag. `EP_NCCL_ROOT_DIR` points DeepEP at that one; installing `nvidia-nccl-cu13` would load a second, differently built NCCL alongside the one torch links.

**Draft pending verification.** The arm64 build is in flight. Once it pushes I'll serve Kimi-K3 on 2x GB300 in one NVLink domain and run the SPEED-Bench qualitative `coding` category (all 80 prompts), checking coherence plus that the MoE backend resolves to a shape the public artifact actually covers. Marking ready once that passes.


abatilo:
## Verification: K3 serves from purely public upstream

Deployed **straight from this image** (`vllm-k3:arm64-public-situ-r1`), not through the infr serving image, so nothing from infr's patch stack is propping it up. 2x GB300, TP8, both pods pinned to one NVLink domain (`DHC2-067-EU-NORTH-04A`).

**Engine state — the backend line is the one that matters:**
```
FLASHINFER_TRTLLM_MXFP4_MXFP8     <- needs MxE4m3_MxE2m1MxE4m3
Auto-selected flashinfer allreduce backend: mnnvl
TRTLLM_RAGGED MLA prefill + FLASHINFER_MLA
GPU KV cache size: 1,942,565
private-pool references: 0        NCCL WARN: 0
```
The public `batched_gemm-3d40263-3e19f0a` artifact ships `siTuGlu` for the `E2m1` and `MxE4m3` families but **not** `Bfloat16_MxE2m1Bfloat16`, so `MXFP4_BF16` would have had no kernels. It resolves to `MXFP4_MXFP8`, which is covered.

**Image checks:** `ActivationType.Situ` present (from upstream), `is_private` gone from `gen_trtllm_gen_fused_moe_sm100_module`, `FLASHINFER_PRIVATE_CUBIN_DIR` unset, `/opt/flashinfer-local-cubins` absent, flashinfer `0.6.16+cu132` installed over vLLM's `0.6.15.post1` pin, and `deep_ep` binds `libnccl.so.2` -> `libnccl.so.2.30.4` matching torch's `(2, 30, 4)`.

**SPEED-Bench qualitative `coding`, all 80 prompts**, concurrency 1, greedy, sequential, `max_tokens 4096`. 80/80 completed, 0 errors, 0 cancelled, 444.8s.

| metric | private-pool baseline (SB8K low-entropy) | this PR (coding) |
|---|---|---|
| inter-token latency | 3.15 ms | **3.12 ms** |
| output tok/s/user | 325.2 | **334.0** |
| E2E tok/s/user | 305.5 | **323.9** |
| cap rate @4096 | 56% | **6.2%** (5/80) |

Output coherent under greedy and temperature 1.0, including correct working code. OSL min 212 / p50 1202 / mean 1610.

## Two notes for the reviewer

**1. `deep_ep` needed a second symlink, now asserted at build time.** The first pass put the NCCL symlink only in the builder stage. `deep_ep` bakes `EP_NCCL_ROOT_DIR` into the wheel and re-globs `<root>/lib` at *import*, so the runtime hit `AssertionError: No libnccl.so found in /usr/lib/` and DeepEP was silently unavailable — visible only as a WARNING. Fixed, and the final stage now asserts `import deep_ep` so a future break fails the build rather than degrading in production. Verified: `deep_ep import OK`.

**2. Pre-existing DeepGEMM divergence, not touched here.** ml-containers main pins `DEEPGEMM_COMMIT=a6b593d2`, but vLLM main's cmake vendors `f5a76426`, and the external wheel shadows vLLM's vendored copy. The result is `deep_gemm.transform_weights_for_mega_moe` without an `activation` kwarg, so **`--moe-backend deep_gemm_mega_moe` is unusable for K3** (it would `TypeError` on `activation="situ"`). It does not affect `auto`/`flashinfer_trtllm`, which is what serves here. I have deliberately left it alone since #202 chose `a6b593d2` for sm12x reasons and reverting it could regress that — flagging rather than deciding.

abatilo:
Marking ready. One transparency note on which artifact each check ran against:

- **Serving + the 80-prompt `coding` run** were on `vllm-k3:arm64-public-situ-r1`, built from the tree before the `deep_ep` symlink fix.
- **`arm64-public-situ-r2`** (`sha256:09a9ddb0`) is built from this exact commit (`c42281a`) and its build asserts `deep_ep import OK`.

The only delta between them is a symlink and an import assertion in the final stage, which cannot affect the K3 flashinfer serving path — so I did not re-run the serving suite on r2. CI will build this commit independently, which is the real gate.

.github/configurations/vllm-tensorizer.yml
```diff
@@ -1,12 +1,10 @@
...
-  # renovate-release marker is what disables it, and no other manager is enabled
-  # for this repo.
+  # Renovate does not track this entry. Having no renovate-release marker is what
+  # disables it, and no other manager is enabled for this repo.
   #
   # TODO(kimi-k3): Re-enable renovate when there's an official release
-  - vllm-commit: '658f2f56e557e8141f8726b4626c51496d78d914'
-    flashinfer-commit: 'v0.6.15.post1'
+  - vllm-commit: '837eae64580c885101ee95b073aafb27a485e7ce'
+    flashinfer-commit: 'a02d94de5796650ead1c6be27b834c3a063bf45d'
```
alexeldeib, line 7: is this the literal head of flashinfer, nice 💀

abatilo:
Waiting for official point release instead

PR #214, opened 2026-08-19
ci: Scope vLLM build caches by ref

alexeldeib:
Let branch builds read the canonical shared cache while writing only to a ref-scoped cache. Queue runs per ref so every pushed SHA builds serially without cross-branch cache races or duplicate ARM contention. This avoids auto-cancellation which I recall Eta preferred, while permitting concurrency groups to prevent same-branch cache collisions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,10 @@ on:
         required: false
         description: "Optional sub-key to append to the image name for build layer caching"
         type: string
+      cache-scope:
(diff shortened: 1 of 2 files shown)


alexeldeib:
mehh this logic is kind of tricky to get the right behavior...using branch alone leads to a cold first build...so ideally you read the shared cache once, then write to branch, then read from branch...mmm

edit: think that kind of works actually?

PR #221, opened 2026-08-27
feat(vllm): compile GLM-5.3 native support on 0.28

sihang-s:
## Summary

- build only the compile-sensitive GLM-5.3-Flash subset into the vLLM 0.28 wheel
- pin FlashInfer to the immutable source for `nightly-v0.6.18-20260819` while keeping CUDA `13.2.1`
- remove the obsolete local FlashInfer routing patch now included upstream
- preserve the existing CoreWeave MNNVL native patch
- leave the Python/runtime files and the two-file narrow block-table fix for downstream runtime layering

## Compiled subset

- `cmake/external_projects/vllm_flash_attn.cmake`
- `csrc/libtorch_stable/cache_kernels.cu`

These files must exist before wheel/native-extension compilation because they pin the flash-attention native dependency and add compiled `head_dim=512` cache-kernel support.

## Source provenance

- CoreWeave base: vLLM `2cf0a6915ce544dc493a0990f2ea38d81601128a`
- Upstream PR: vllm-project/vllm#53906
- Reviewed upstream head: `0f82208cb1d01e723bb25b607217a1adfb0cfc30`
- Original feature commit: `933876c388fb129ad82590660e6506614559cb86`
- FlashInfer: `ee3fda10e0b9d82b71bda534bc2954d3176e6ded` (`nightly-v0.6.18-20260819`)
- FlashInfer no-group routing fix: `08ddfbcd2e89b2f4b68391825817909e30d445e2` (already included in the pinned source)
- flash-attention: `vllm-project/flash-attention@4eb29359a51b4fe67c8c0f0c5e931236971aa492`
- Compile patch SHA256: `00e84751bd79e3140fc4344223d9997548605d47ba66f4ee51d108520831d762`

## Existing native patch consolidation

- moved the MNNVL Lamport mailbox fix into the numbered `vllm-patches/` stack
- refreshed it to the final upstream source hunk from vLLM #53000 at `a047e2543da570a64d1bbfeac4fe44eff3e87a81`
- retained it because the pinned `v0.28.0` source predates the fix; the first containing release tag is `v0.28.1rc0`
- MNNVL patch SHA256: `d4e98e4b302ecdaf463120b7ca1b1f567c18525b742ae83d2d9a445c042173e9`

## Compatibility contract

The downstream runtime patch must use the same vLLM base and reviewed upstream head. Python/runtime-only upstream deltas update only the downstream layer; CMake, CUDA/C++, native-binding, native-dependency, or FlashInfer changes require a new compiled image and digest.

## Validation

- compile-only patch applies cleanly to the exact vLLM 0.28 source pin
- existing MNNVL patch applies after the compiled subset
- Docker buildx checks pass for `linux/amd64` and `linux/arm64` with no warnings
- `vllm-downloader` and `flashinfer-downloader` stages pass for `linux/arm64`
- full amd64/arm64 wheel and image builds are delegated to this PR's CI

## Follow-up

After CI publishes the new immutable multi-arch digest, downstream consumers can pin this compiled image and layer the Python/runtime GLM backport.


vllm-tensorizer/vllm-patches/0002-guard-narrow-block-table-slot-mapping.patch
```diff
@@ -0,0 +1,58 @@
+diff --git a/vllm/v1/worker/block_table.py b/vllm/v1/worker/block_table.py
+index 85228364af..f7372c2d46 100644
+--- a/vllm/v1/worker/block_table.py
```
nehashah-eng: I am guessing all python files would not require compilation . So can move to infr/ 

I believe only part of  0001 requires compilation

vllm-tensorizer/vllm-patches/0001-backport-glm-5-3-flash.patch
```diff
@@ -0,0 +1,11980 @@
...
+                                CALL_GATHER_CACHE_320);
+diff --git a/vllm/compilation/passes/fusion/allreduce_rms_fusion.py b/vllm/compilation/passes/fusion/allreduce_rms_fusion.py
+index fdfc5b8be4..7660686529 100644
+--- a/vllm/compilation/passes/fusion/allreduce_rms_fusion.py
++++ b/vllm/compilation/passes/fusion/allreduce_rms_fusion.py
+@@ -257,9 +257,11 @@ if flashinfer_comm is not None:
+             residual_out = allreduce_in
+ 
+         layout_code = None
+-        # vLLM quant patterns use swizzled scale-factor layout. Non-quant
```
nehashah-eng: Like you can move all these to infr/ and just keep the cmake stuff here ..

vllm-tensorizer/Dockerfile
```diff
@@ -86,6 +86,14 @@ RUN git clone --filter=tree:0 --no-single-branch --no-checkout \
...
       --depth 1 --filter=tree:0
+
+# Apply source backports before building the vLLM wheel. GLM-5.3 changes
+# include CMake, CUDA, and external-kernel revisions, so applying them only to
+# site-packages in the downstream runtime image would leave stale binaries.
+COPY vllm-patches/ /git/vllm-patches/
+RUN git -C vllm apply -v --stat --apply /git/vllm-patches/*.patch && \
+    rm -rf /git/vllm-patches
+
 COPY vllm-mnnvl-lamport-multimem-store.patch /git/patch
```
alexeldeib: can we consolidate this into above folder now that there are more, should actually be able to remove it in a new-ish release but not sure it made it into 0.28

nehashah-eng:
@alexeldeib  also has other patch here - https://github.com/coreweave/ml-containers/tree/main/vllm-tensorizer ... maybe we should use all the patches in the same way.

vllm-tensorizer/Dockerfile
sihang-s: Done. I moved the MNNVL fix into the numbered vllm-patches stack as 0002 and removed the separate Dockerfile apply step. The pinned v0.28.0 source (2cf0a691) does not contain the upstream fix; it landed at a047e2543d in #53000 and first appears in v0.28.1rc0. I also refreshed the patch to that final upstream source hunk, so it can be dropped when the base moves to a containing revision.

vllm-tensorizer/vllm-patches/0001-backport-glm-5-3-flash.patch
```diff
@@ -0,0 +1,76 @@
+diff --git a/cmake/external_projects/flashmla.cmake b/cmake/external_projects/flashmla.cmake
```
nehashah-eng: Also for the sake of preserving the numbering lets call this 0003-

nehashah-eng: I am guessing there was a 0001 before which was deleted.

vllm-tensorizer/Dockerfile
```diff
@@ -347,6 +352,14 @@ RUN --mount=type=secret,id=s3_access_key_id,env=AWS_ACCESS_KEY_ID \
 
 FROM ${FINAL_BASE_IMAGE} AS base
 
+ARG VLLM_COMMIT
+ARG FLASHINFER_COMMIT
+ARG VLLM_GLM53_COMPILED_REF=142062f13d16bed254b5d97cc3d371fbd4f7790a
```
alexeldeib: What is this value?

alexeldeib: The post patch sha or something?

sihang-s: upstream vLLM PR #53906 head from which the three compile-sensitive hunks were derived, not a post-patch SHA.
Does not affect the build

alexeldeib (dismissed):
seems okay assuming it works 🤞 

Let's strip out the 2 forks as soon as convenient in follow up when upstream is ready

.github/configurations/vllm-tensorizer.yml
```diff
@@ -1,7 +1,7 @@
 include:
   # renovate-release: datasource=github-releases depName=vllm-project/vllm currentValue=v0.28.0 versioning=pep440
   - vllm-commit: '2cf0a6915ce544dc493a0990f2ea38d81601128a'
-    flashinfer-commit: 'v0.6.16.post3'
+    flashinfer-commit: 'e62941a1da605fb9b3c8c50b23c9720df12cf6b4'
```
nehashah-eng, line 4: Just to confirm this is a release right, not an arbitrary commit ?

nehashah-eng (approved):

.github/configurations/vllm-tensorizer.yml
sihang-s: It's from a RC tag. Pinging commit SHA here for immutable build.

sihang-s: v0.6.18rc10

PR #232, opened 2026-09-11
fix(sglang): expose system NCCL to DeepEP

nehashah-eng:
## Summary

- expose the apt-installed NCCL libraries and headers through a canonical `/opt/nccl` prefix
- set `EP_NCCL_ROOT_DIR=/opt/nccl` for sgl-deep-ep discovery
- add a build-time regression check using DeepEP find_nccl_root logic

## Problem

The SGLang 0.5.19 Dynamo worker exits during `deep_ep` import with:

```
AssertionError: Cannot find package: nccl
```

NCCL 2.30.4 and `libnccl.so` are already installed by the CUDA base through apt. DeepEP 2.1.0 discovers NCCL using an `nvidia-nccl` Python distribution or an explicit root, so it cannot locate the system package without this prefix.

The fix reuses the existing system NCCL installation and does not add a second NCCL version.

## Validation

- `bash -n sglang/install.bash`
- `git diff --check`
- exercised the exact installed DeepEP discovery implementation in the affected image using the proposed prefix
- verified discovery of `libnccl.so`, `libnccl.so.2`, `libnccl.so.2.30.4`, and `nccl.h`

A full top-level DeepEP import requires a GPU and therefore cannot run during the container build; the regression check invokes its discovery module directly without bypassing the code that originally failed.

### DIFF
--- a/sglang/Dockerfile
+++ b/sglang/Dockerfile
@@ -53,10 +53,20 @@ RUN apt-get -qq update && \
     apt-get -q install --no-install-recommends -y \
       libnvshmem3-cuda-13 \
       libssl-dev && \
+    NCCL_LIB_DIR="/usr/lib/$(dpkg-architecture -qDEB_HOST_MULTIARCH)" && \
     NVSHMEM_LIB_DIR="/usr/lib/$(dpkg-architecture -qDEB_HOST_MULTIARCH)/nvshmem/13" && \
+    mkdir -p /opt/nccl && \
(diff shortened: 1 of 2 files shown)


JustinPerlman (approved):

alexeldeib (approved):
python packages seem comically bad at differentiating between system and python packages, sigh

one day I will learn how they figure this out and why they don't just use site_packages followed by LD_LIBRARY_PATH

PR #238, opened 2026-09-18
feat: Move the patches to infr

gshtrasberg:
Pending validation in https://github.com/coreweave/infr/tree/greg/dsv41-patch


gshtrasberg:
omg why is everything getting rebuilt

